PR #27, opened 2024-10-08
Doug/fix run issue

dacoburn:
Fixed an issue where the scan was not running for the default branch in Github

### DIFF
--- a/socketsecurity/__init__.py
+++ b/socketsecurity/__init__.py
@@ -1,2 +1,2 @@
 __author__ = 'socket.dev'
-__version__ = '1.0.26'
+__version__ = '1.0.30'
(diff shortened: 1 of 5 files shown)


socketsecurity/__init__.py
```diff
@@ -1,2 +1,2 @@
 __author__ = 'socket.dev'
-__version__ = '1.0.30'
+__version__ = '1.0.31'
```
reberhardt7, line 2: Please add a PR title or description explaining what the problem is, even if it's just a copy/paste error message

reberhardt7 (approved):

socketsecurity/__init__.py
dacoburn: Done!

PR #28, opened 2024-10-18
Added support for CycloneDX and SPDX manifest files to the CLI. Also …

dacoburn:
…improved build script


obarrera (approved):

mikolalysenko (approved):
Seems fine, but we should probably pull these glob patterns from the API instead of hard coding them.  @bcomnes do we have an API for this?

flowstate:
> Seems fine, but we should probably pull these glob patterns from the API instead of hard coding them. @bcomnes do we have an API for this?

If we do have an API for this, or even just an established pattern for generating the glob patterns in the JS sdk, I can incorporate that into the changes I'm working on right now. 

I'm working through the cli, replacing all the direct API calls with the associated methods from the python SDK.

PR #29, opened 2024-11-01
migrate from direct API usage to SDK

flowstate:
## Changes

- Migrated all the direct calls to the API to instead use the python SDK
- Slight refactor of the Core class to explicitly make it a static class instead of mixing static and instance variables / state
- Added the dev dependency Ruff, a modern, more performant language server and linter
- Added python-dotenv and some config for setting up a .env.local file for setting env vars

I've attempted to thoroughly test these changes, but given how new I am to the codebase and company, I'd love a sanity check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ markdown_security_temp.md
 test.py
 *.cpython-312.pyc`
 file_generator.py
+.env.local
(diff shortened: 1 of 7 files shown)


dacoburn (approved):

flowstate:
@SocketSecurity ignore pypi/sca@0.0.1 pypi/oss@0.1.3

ignoring the Staging Alerts due to the incorrect parsing of the pyproject.toml

PR #34, opened 2024-12-05
added pipfile.lock to socket globs

flowstate:


flowstate:
need to update the version checker to remove any old comments if the version has been fixed

dacoburn (approved):

PR #36, opened 2025-01-14
comparing file paths on manifest files to dedupe

flowstate:
Our case-insensitive file search led to us sending multiple copies of manifest files when we created a full scan. Now we are tracking seen file paths, so we only send one copy of each unique manifest


flowstate:
Closing since Doug's better solution is live and verified working

PR #38, opened 2025-01-17
Fixed the logic for case insensitive to not do multiple globs but sti…

dacoburn:
…ll keep the correct path names


flowstate (approved):
LGTM!

PR #43, opened 2025-01-24
Added SARIF support to the CLI

obarrera:


socketsecurity/socketcli.py
```diff
@@ -188,6 +196,25 @@ def output_console_comments(diff_report: Diff, sbom_file_name: str = None) -> No
         else:
             log.info("No New Security issues detected by Socket Security")
 
+def output_console_sarif(diff_report: Diff, sbom_file_name: str = None) -> None:
```
dacoburn, line 199: @obarrera is the intention to replace printing out to the console for JSON or table format or to have it save to a file? Right now this would replace JSON/TABLE output and print out in sarig format.  The other option is still do the normal output and save a file for sarif. This definitely works but I'm not sure what is better as I haven't played with sending to Github.

dacoburn (approved):
Overall looks good, just have the one comment/question. Also update the following to the next version (1.0.43)

https://github.com/SocketDev/socket-python-cli/blob/8b04089c2d2c658b636b02e9fb2940398edd3fd4/socketsecurity/__init__.py#L2

socketsecurity/socketcli.py
obarrera: Right now, this will just print the sarif to the console, but we can modify the behavior later if needed.

PR #60, opened 2025-03-07
Now using supported files SDK endpoint for file globs

flowstate:


dacoburn (changes requested):
Need to set the exclusion before this can be pushed. Otherwise it'll cause a regression

dacoburn (approved):

PR #61, opened 2025-03-10
Eric/use types and logging

flowstate:
- Fixes the debug logging (don't know the root cause yet, but this solves it)
- Correctly handles success and failure of create repo SDK call if there isn't an existing repo

We **thought** we need to add `use_types=True`, but there are no types for the post method in the SDK. 

The actual problem was that the SDK's repos.post method has the following code:

```python
    def post(self, org_slug: str, **kwargs) -> dict:
        params = {}
        if kwargs:
            for key, val in kwargs.items():
                params[key] = val
        if len(params) == 0:
            return {}

        path = "orgs/" + org_slug + "/repos"
        payload = json.dumps(params)
        response = self.api.do_request(path=path, method="POST", payload=payload)

        if response.status_code == 201:
            return response.json()

        error_message = response.json().get("error", {}).get("message", "Unknown error")
        log.error(f"Error creating repository: {response.status_code}, message: {error_message}")
        return {}
```

So the fix was to check for success by checking if the response had content or not. 

I've also wrapped this create call in its own try/except, and if it fails we log it and exit with code 2. 

So now if both calls fail there won't be an "unexpected error while running the CLI"

======

Proof that the logging issue is gone:

with debug enabled:
```shell
$> socketcli --repo test-large-repo --branch "test-cli-v2-perf" --pr_number 12 --target_path ../test-large-repo --timeout 1200 --enable-debug --exclude-license-details --ignore-commit-files
DEBUG:socketcli:Debug logging enabled
DEBUG:socketcli:loaded socket_config
DEBUG:socketcli:loaded client
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.socket.dev:443
DEBUG:urllib3.connectionpool:https://api.socket.dev:443 "GET /v0/organizations HTTP/11" 200 None
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.socket.dev:443
DEBUG:urllib3.connectionpool:https://api.socket.dev:443 "GET /v0/orgs/socketdev-demo/settings/security-policy 
...
INFO:socketdev:Total time to create new full scan: 5.81
INFO:socketdev:No head scan found. New scan ID: 089908d7-c2c7-40b3-aa65-791e35942cb5
INFO:socketcli:No issues found
```

without debug:

```shell
$> socketcli --repo test-large-repo --branch "test-cli-v2-perf" --pr_number 12 --target_path ../test-large-repo --timeout 1200  --exclude-license-details --ignore-commit-files 

INFO:socketcli:API Mode
INFO:socketdev:Total time to create new full scan: 17.40
INFO:socketdev:No head scan found. New scan ID: 75f0f2d1-8f45-4d98-b199-5dc9ac11be3f
INFO:socketcli:No issues found

```

### DIFF
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -394,11 +394,10 @@ class Core:
             if not response.success:
                 log.error(f"Failed to get repository: {response.status}")
                 log.error(response.message)
-                # raise Exception(f"Failed to get repository info: {response.status}, message: {response.message}")
         except APIFailure:
             log.warning(f"Failed to get repository {repo_slug}, attempting to create it")
             try:
-                # Remove use_types=True since post() doesn't support it
+
                 create_response = self.sdk.repos.post(self.config.org_slug, name=repo_slug, default_branch=default_branch)
 
                 # Check if the response is empty (failure) or has content (success)
@@ -406,11 +405,9 @@ class Core:
                     log.error("Failed to create repository: empty response")
                     raise Exception("Failed to create repository: empty response")
                 else:
-                    # If we got here, create_response is a dictionary with the repository data
-                    return create_response  # This is already the repository data
+                    return create_response
 
             except APIFailure as e:
-                # Handle API failures from the post request
                 log.error(f"API failure while creating repository: {e}")
                 sys.exit(2) # Exit here with code 2. Code 1 indicates a successfully-detected security issue.
 
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -397,15 +397,23 @@ class Core:
                 # raise Exception(f"Failed to get repository info: {response.status}, message: {response.message}")
         except APIFailure:
             log.warning(f"Failed to get repository {repo_slug}, attempting to create it")
-            create_response = self.sdk.repos.post(self.config.org_slug, name=repo_slug, default_branch=default_branch, use_types=True)
-            if not create_response.success:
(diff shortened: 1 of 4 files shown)


socketsecurity/core/__init__.py
```diff
@@ -394,18 +394,23 @@ def get_repo_info(self, repo_slug: str, default_branch: str = "socket-default-br
...
-                log.error(f"Failed to create repository: {create_response.status}")
-                log.error(create_response.message)
-                raise Exception(
-                    f"Failed to create repository: {create_response.status}, message: {create_response.message}"
-                )
-            else:
-                return create_response.data
+            try:
+
+                create_response = self.sdk.repos.post(self.config.org_slug, name=repo_slug, default_branch=default_branch)
```
dacoburn, line 401: I don't see the `use_types=True`?

socketsecurity/core/logging.py
```diff
@@ -23,10 +24,24 @@ def initialize_logging(
     cli_logger = logging.getLogger(cli_logger_name)
     cli_logger.setLevel(level)
 
+    # Explicitly set urllib3 logger to WARNING to prevent debug messages
+    # when not in debug mode
+    urllib3_logger = logging.getLogger("urllib3")
```
dacoburn, line 29: In `2.0.9` we didn't need to do this, so did we figure out what it was about `2.0.11` that triggered this messages showing?

dacoburn (changes requested):

dacoburn (approved):

PR #93, opened 2025-06-30
Add guardrail in case manifest file does not exist

dacoburn:
## Root Cause

Logic for only creating the license file if a diff scan was ran was incorrect

## Fix

Changed logic for detection

## Public Changelog

N/A


pvdz (approved):

socketsecurity/core/__init__.py
```diff
@@ -860,9 +860,10 @@ def get_source_data(package: Package, packages: dict) -> list:
             manifests = ""
             if not hasattr(package, "manifestFiles"):
                 return introduced_by
-            for manifest_data in package.manifestFiles:
-                manifest_file = manifest_data.get("file")
-                manifests += f"{manifest_file};"
+            if hasattr(package, "manifestFiles"):
```
Planeshifter, line 863: This could be simplified to `else:`, no?

socketsecurity/socketcli.py
```diff
@@ -260,7 +260,7 @@ def main_code():
         output_handler.handle_output(diff)
 
         # Handle license generation
-    if diff is not None and diff.id != "no_diff_id" and config.generate_license:
+    if should_skip_scan and diff.id != "no_diff_id" and config.generate_license:
```
Planeshifter, line 263: Is this correct? Or did you mean to check for whether NOT should_skip_scan here?

PR #94, opened 2025-06-30
Revert "Add guardrail in case manifest file does not exist"

dacoburn:
Reverts SocketDev/socket-python-cli#93


pvdz (approved):
huh? ok...

PR #101, opened 2025-07-23
Refactor file detection and scanning logic to fix commit file handling

dacoburn:
## Root Cause

The files logic for detecting if there were changed manifest files had been broken over time. At this point it was completely decoupled and in most instances things didn't work correctly without `--ignore-commit-files`

## Fix

- Fix file argument parsing to handle list, string, and JSON formats more robustly
- Clarify git repository detection and file selection logic with better separation of concerns
- Add force_api_mode to handle cases where no supported manifest files are found
- Replace ambiguous should_skip_scan logic with clearer file detection flow
- Add create_full_scan_with_report_url method to Core for API-mode scanning
- Improve logging messages and remove unused code (get_all_scores method)
- Ensure consistent diff object initialization and ID handling
- Automatically enable disable_blocking when no supported files are detected

## Public Changelog

• Fixed file detection logic that was preventing proper scanning of changed manifest files
• Improved handling of different file input formats (list, string, JSON) for better reliability
• Enhanced automatic detection of git repository changes and manifest file filtering
• Fixed cases where scans would incorrectly skip when manifest files were present in commits

### DIFF
--- a/socketsecurity/__init__.py
+++ b/socketsecurity/__init__.py
@@ -1,2 +1,2 @@
 __author__ = 'socket.dev'
-__version__ = '2.1.21'
+__version__ = '2.1.22'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "socketsecurity"
-version = "2.1.21"
+version = "2.1.22"
 requires-python = ">= 3.10"
 license = {"file" = "LICENSE"}
 dependencies = [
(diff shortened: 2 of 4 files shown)


socketsecurity/__init__.py
```diff
@@ -1,2 +1,2 @@
 __author__ = 'socket.dev'
-__version__ = '2.1.21'
```
ahmadnassri, line 2: can we use the version in the HTTP user-agent(s)?

philidem (approved):

PR #178, opened 2026-04-02
feat(ignore): adding event logger for ignored comments

billxinli:
- Emits events to `POST /v0/orgs/{slug}/telemetry` when users suppress alerts via `@SocketSecurity ignore` PR/MR comments
- Uses `eyes` emoji reaction as a dedup marker — only unprocessed comments (without `eyes`) trigger events; after sending, the `eyes` reaction is added
- Per-comment iteration ensures the correct comment author is recorded as `sender_name`/`sender_id`
- Supports both GitHub (inline `reactions.eyes` from API response) and GitLab (lazy `has_eyes_reaction()` API call per comment, best-effort `post_eyes_reaction()` via Award Emoji API)
- Events are sent one at a time (fire-and-forget per event, continues on failure)
- Comment flow uses `artifact_input` (raw user text) since the ignore command is user input and may not be a valid PURL
- Push flow uses `artifact_purl` from actual alert objects with `alert_action` derived from the alert's resolved policy action
- Fixes pre-existing case-sensitivity bug in `get_ignore_options` — `@socketSecurity ignore` (lowercase s) now works

### `socketsecurity/core/cli_client.py`
- Added `post_telemetry_events()` — sends events individually to `POST /v0/orgs/{slug}/telemetry`

### `socketsecurity/core/scm/github.py`
- Added `is_commenter_authorized()` — checks collaborator permission level (admin/write)
- Added `post_eyes_reaction()` — posts `eyes` reaction to mark comments as processed
- Added `post_negative_reaction()` — posts `-1` reaction for unauthorized users
- Updated `handle_ignore_reactions()` — only adds `+1` for authorized users, `-1` for unauthorized

### `socketsecurity/core/scm/gitlab.py`
- Added `is_commenter_authorized()` — checks project member access level (Developer+)
- Added `has_eyes_reaction()` — best-effort check for `eyes` award emoji on MR notes
- Added `post_eyes_reaction()` — best-effort add `eyes` award emoji with `Content-Type: application/json`

### `socketsecurity/core/scm_comments.py`
- Made `get_ignore_options()` case-insensitive (fixes `@socketSecurity` vs `@SocketSecurity`)

### `socketsecurity/socketcli.py`
- Added `_is_commenter_authorized()` — checks user permissions with caching per user
- Added `_is_unprocessed()` — checks inline `reactions.eyes`, falls back to `scm.has_eyes_reaction()`, then checks authorization
- Added `_filter_authorized_ignore_comments()` — filters ignore comments to authorized users only, applied before suppression and telemetry in both flows
- Comment flow: emits events per unprocessed ignore comment with author attribution
- Push flow: matches ignored alerts back to individual comments, derives `alert_action` from alert's resolved policy flags

### Tests
- `tests/unit/test_client.py` — 2 tests: individual event sending, continues on failure

## Event attributes

```json
{
  "event_kind": "user-action",
  "client_action": "ignore",
  "alert_action": "error",
  "event_id": "<uuid>",
  "event_sender_created_at": "<iso8601>",
  "vcs_provider": "github|gitlab",
  "owner": "<repo_owner>",
  "repo": "<owner/repo>",
  "pr_number": 123,
  "ignore_all": true|false,
  "sender_name": "<comment_author_login>",
  "sender_id": "<comment_author_id>",
  "artifact_input": "<raw_user_text>" (comment flow),
  "artifact_purl": "<valid_purl>" (push flow)
}
```

## Public Changelog

N/A

### DIFF
--- a/socketsecurity/socketcli.py
+++ b/socketsecurity/socketcli.py
@@ -4,6 +4,8 @@ import sys
 import traceback
 import shutil
 import warnings
+from datetime import datetime, timezone
+from uuid import uuid4
 
 from dotenv import load_dotenv
 from git import InvalidGitRepositoryError, NoSuchPathError
@@ -478,6 +480,17 @@ def main_code():
 
     # Handle SCM-specific flows
     log.debug(f"Flow decision: scm={scm is not None}, force_diff_mode={force_diff_mode}, force_api_mode={force_api_mode}, enable_diff={config.enable_diff}")
+
+    def _is_unprocessed(c):
+        """Check if an ignore comment has not yet been marked with 'eyes' reaction.
+        For GitHub, reactions.eyes is already in the comment response (no extra call).
+        For GitLab, has_eyes_reaction() makes a lazy API call per comment."""
+        if getattr(c, "reactions", {}).get("eyes"):
+            return False
+        if hasattr(scm, "has_eyes_reaction") and scm.has_eyes_reaction(c.id):
+            return False
+        return True
+
     if scm is not None and scm.check_event_type() == "comment":
         # FIXME: This entire flow should be a separate command called "filter_ignored_alerts_in_comments"
         # It's not related to scanning or diff generation - it just:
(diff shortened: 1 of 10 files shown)


billxinli:
bugbot run

bmeck (approved):
seems ok, would slightly pref v1 events but this is fine since v1 would be a slightly bigger change

socketsecurity/socketcli.py
```diff
@@ -486,10 +499,52 @@ def main_code():
...
+                try:
+                    events = []
+                    for c in unprocessed:
+                        single = {"ignore": [c]}
+                        ignore_all, ignore_commands = Comments.get_ignore_options(single)
+                        user = getattr(c, "user", None) or getattr(c, "author", None) or {}
+                        now = datetime.now(timezone.utc).isoformat()
+                        shared_fields = {
+                            "event_kind": "user-action",
+                            "client_action": "ignore_alerts",
```
bmeck: action fields should match ResolvedIssueAction, and this is missing the current action from policy (alert_action)

```
enum ResolvedIssueActionEnum {
  error = 'error',
  warn = 'warn',
  monitor = 'monitor',
  ignore = 'ignore',
}
```

socketsecurity/socketcli.py
```diff
@@ -478,6 +480,17 @@ def main_code():
 
     # Handle SCM-specific flows
     log.debug(f"Flow decision: scm={scm is not None}, force_diff_mode={force_diff_mode}, force_api_mode={force_api_mode}, enable_diff={config.enable_diff}")
+
+    def _is_unprocessed(c):
+        """Check if an ignore comment has not yet been marked with '+1' reaction.
```
jhiesey, line 485: Is this the best way of marking comments processed? Seems a little easy for other human users to mess up. It's fine as a best-effort approach though

socketsecurity/socketcli.py
```diff
@@ -504,9 +555,76 @@ def main_code():
             # FIXME: this overwrites diff.new_alerts, which was previously populated by Core.create_issue_alerts
             if not config.disable_ignore:
                 log.debug("Removing comment alerts")
+                alerts_before = list(diff.new_alerts)
                 diff.new_alerts = Comments.remove_alerts(comments, diff.new_alerts)
```
jhiesey, line 559: Why does this use `Comments.remove_alerts` but the above code path uses `scm.remove_comment_alerts(comments)`?

jhiesey (approved):
A couple questions but otherwise looks good!

socketsecurity/socketcli.py
billxinli: This is mainly used to dedup the number of events logged.

first `ignore foo` message will cause the workflow to rerun with the cli, and it will log foo is being ignored

second `ignore bar` message will cause foo and bar to be logged, the +1 emoji attempts to deduplicate these logs, however, at the same time, it also means the user could immediately add the +1 emoji

socketsecurity/socketcli.py
billxinli: My understanding of this is that there are two different flows.
Comment flow: Security comment already exists on the PR. User posts @SocketSecurity ignore foo. The CLI re-processes the existing comment to remove foo rows.

Push flow: A new push happened. The CLI is about to create a new security comment. It filters alerts before generating the comment so ignored packages never appear.

one `scm.remove_comment_alerts` actually modify the comment as part of the PR/MR, while `Comments.remove_alerts` is just in memory filter

jhiesey: Ah ok, makes sense

socketsecurity/socketcli.py
jhiesey: Got it

PR #193, opened 2026-04-24
fix: bump Socket SDK version, handle missing diff scores

### LINKED ISSUE #78: ValueError: 'other' is not a valid SocketCategory when parsing stream_diff response
## Summary

`socketdev.fullscans.stream_diff` crashes with `ValueError: 'other' is not a valid SocketCategory` when the API returns an alert whose `category` is `"other"`. The `SocketCategory` enum in the SDK does not include `"other"`, so `SocketAlert.from_dict` blows up during deserialization.

## Stack trace

Observed via `socketsecurity==2.2.79` (which pulls `socketdev==3.0.32`) running in CI:

```
2026-04-23 19:14:07,719: Error getting diff report: 'other' is not a valid SocketCategory
  File ".../socketsecurity/core/__init__.py", line 876, in get_added_and_removed_packages
    self.sdk.fullscans.stream_diff
  File ".../socketdev/fullscans/__init__.py", line 873, in stream_diff
    return StreamDiffResponse.from_dict({"success": True, "status": 200, "data": result})
  File ".../socketdev/fullscans/__init__.py", line 640, in from_dict
    data=FullScanDiffReport.from_dict(data_value) if data_value else None,
  File ".../socketdev/fullscans/__init__.py", line 616, in from_dict
    artifacts=DiffArtifacts.from_dict(data["artifacts"]),
  File ".../socketdev/fullscans/__init__.py", line 556, in from_dict
    added=[DiffArtifact.from_dict(a) for a in data["added"]],
  File ".../socketdev/fullscans/__init__.py", line 517, in from_dict
    alerts=[SocketAlert.from_dict(alert) for alert in data.get("alerts", [])],
  File ".../socketdev/fullscans/__init__.py", line 450, in from_dict
    category=SocketCategory(data["category"]),
  File ".../enum.py", line 751, in __call__
    return cls.__new__(cls, value)
  File ".../enum.py", line 1170, in __new__
    raise ve_exc
ValueError: 'other' is not a valid SocketCategory
```

## Root cause

`SocketCategory` at current `main` HEAD (verified via `gh api repos/SocketDev/socket-sdk-python/contents/socketdev/fullscans/__init__.py`) defines only:

```python
class SocketCategory(str, Enum):
    SUPPLY_CHAIN_RISK = "supplyChainRisk"
    QUALITY = "quality"
    MAINTENANCE = "maintenance"
    VULNERABILITY = "vulnerability"
    LICENSE = "license"
    MISCELLANEOUS = "miscellaneous"
```

The Socket.dev API is now returning `"other"` as an alert category, which breaks any consumer of the SDK on any PR whose diff happens to include a package with such an alert.

## Impact

Every PR in our repo that bumps certain Android Compose dependencies (e.g. `androidx.compose:compose-bom` to `2026.04.01`, `androidx.compose.ui:ui-tooling` to `1.11.0`) now fails its `socketsecurity` CI step with this error, and the job skips downstream gates — effectively a hard block on those PRs. Other PRs pass because their diffs don't include whichever package carries the `"other"`-category alert. Reruns fail deterministically with the same `ValueError`.

## Suggested fix

Either:

1. **Lenient deserialization** — fall back to `MISCELLANEOUS` (or a new `UNKNOWN`) when the API emits a category the SDK doesn't recognize. This future-proofs against any new category Socket.dev adds server-side before the SDK catches up.

   ```python
   try:
       category = SocketCategory(data["category"])
   except ValueError:
       category = SocketCategory.MISCELLANEOUS  # or UNKNOWN
   ```

2. **Add `OTHER = "other"` to the enum** — works for this specific case but leaves the SDK brittle to the next server-side category addition.

Option 1 is what we'd ask for — the current behavior makes the SDK a synchronization liability between Socket.dev's server changes and SDK releases.

## Environment

- `socketsecurity==2.2.79`
- `socketdev==3.0.32` (transitively; also latest on PyPI)
- Python 3.12.13, Ubuntu 24.04 GitHub Actions runner

Happy to send a PR for the lenient-deserialization fix if that's the preferred approach.

lelia:
## Summary

Bump the CLI to `socketdev>=3.0.33,<4.0.0` to pick up [the SDK fix](https://github.com/SocketDev/socket-sdk-python/pull/79) for unknown alert categories, and harden dependency overview rendering so PR comments no longer crash when diff artifacts have missing or null score data.

## Problem

1. The SDK could crash while deserializing diff alerts when the API returned an unknown category like `"other"`. That is fixed in `socketdev 3.0.33`, so the CLI needs to consume that release.
2. The CLI could crash while generating the dependency overview comment when a diff artifact had `score` / `scores` set to `None`. The comment renderer assumed `added.scores` was always a dict and called `.get(...)` on it.

## Fix

- bump the CLI dependency to `socketdev>=3.0.33,<4.0.0`
- bump CLI version to `2.2.86` to prepare for new release
- normalize diff artifacts with `score=None` to an empty score map in the CLI model layer
- make dependency overview badge rendering tolerate missing, null, or partial score data
- default missing badge values to a valid 100% fallback instead of producing invalid badge URLs

## Test plan

- [x] `uv run --refresh --extra test pytest tests/unit/test_dependency_overview.py tests/core/test_package_and_alerts.py -q`

Added regression coverage for:
- `Package.from_diff_artifact(...)` normalizing `score=None`
- dependency overview comment generation succeeding when package scores are missing or partially null
- valid fallback badge output for missing scores

Also fixes: https://github.com/SocketDev/socket-sdk-python/issues/78

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "socketsecurity"
-version = "2.2.85"
+version = "2.2.86"
 requires-python = ">= 3.11"
 license = {"file" = "LICENSE"}
 dependencies = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
     'GitPython',
     'packaging',
     'python-dotenv',
-    "socketdev>=3.0.32,<4.0.0",
+    "socketdev>=3.0.33,<4.0.0",
     "bs4>=0.0.2",
     "markdown>=3.10",
 ]
--- a/socketsecurity/__init__.py
+++ b/socketsecurity/__init__.py
@@ -1,3 +1,3 @@
 __author__ = 'socket.dev'
-__version__ = '2.2.85'
+__version__ = '2.2.86'
 USER_AGENT = f'SocketPythonCLI/{__version__}'
--- a/socketsecurity/core/classes.py
+++ b/socketsecurity/core/classes.py
@@ -207,7 +207,7 @@ class Package():
                 name=data["name"],
                 version=data["version"],
                 type=data["type"],
-                score=data.get("score", data.get("scores", {})),
+                score=data.get("score") or data.get("scores") or {},
                 alerts=data.get("alerts", []),
                 author=data.get("author", []),
                 size=data.get("size"),
@@ -236,7 +236,7 @@ class Package():
             name=data["name"],
             version=data["version"],
             type=data["type"],
-            score=data.get("score", data.get("scores", {})),
+            score=data.get("score") or data.get("scores") or {},
             alerts=data.get("alerts", []),
             author=data.get("author", []),
             size=data.get("size"),
@@ -448,6 +448,8 @@ class Purl:
             self.capabilities = []
         if not hasattr(self, "is_new"):
             self.is_new = False
+        if not hasattr(self, "scores") or self.scores is None:
+            self.scores = {}
         self.author_url = Purl.generate_author_data(self.author, self.ecosystem)
 
     @staticmethod
--- a/socketsecurity/core/messages.py
+++ b/socketsecurity/core/messages.py
@@ -1179,12 +1179,27 @@ class Messages:
                 score_percent = int(score * 100)  # Convert to integer percentage
                 return f"[![{score_percent}](https://github-app-statics.socket.dev/score-{score_percent}.svg)]({added.url})"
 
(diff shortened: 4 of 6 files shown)


johnsonch (approved):
Thank you for doing this!

dc-larsen (commented):
LGTM

dc-larsen (commented):
LGTM

dc-larsen (commented):
LGTM

philgran (approved):

dc-larsen (approved):
LGTM

PR #201, opened 2026-05-12
feat: stream CLI log transcripts and run status to Socket backend

barslev:
Adds an opt-in streaming log channel between the Python CLI and the
Socket backend so the CLI run's terminal status (`in_progress` /
`success` / `failure` / `cancelled`) and a transcript of its own log
output are visible in the Socket admin views when the user opts in
with `--upload-logs`. The Socket backend may also force-enable
streaming for specific orgs regardless of the flag.

## Why?

The Socket backend currently has no visibility into what happens
inside a CLI invocation — there's no record of whether a scan ran
to completion or what was logged along the way. This PR opens a
bounded, opt-in side-channel that uploads the CLI's own log records
to the backend and reports the run's terminal status, without
changing any existing CLI request on the wire.

### New modules
- `socketsecurity/core/cli_run.py` — `register_cli_run` and
  `finalize_cli_run` lifecycle helpers, both best-effort.
  `register_cli_run` always calls `POST /v0/python-cli-runs` with the
  user's `share_logs` choice; the server responds with
  `log_streaming_enabled` and a nullable `run_id`. The CLI gates the
  rest of the lifecycle on `log_streaming_enabled` (the server can
  force-enable for an org even when the client didn't ask).
  `finalize_cli_run` accepts an optional `report_run_id` that links
  the run to the full-scan it produced.
- `socketsecurity/core/log_uploader.py` — `BatchedLogUploader`
  (daemon-thread flusher, 5s flush interval, swallows all network
  errors at-most-once, skips empty buffers, drains on shutdown) plus
  `UploadingLogHandler` — a `logging.Handler` that routes records
  to the uploader. Includes a thread-local recursion guard so the
  uploader's own request-error logs aren't re-enqueued mid-flush.
- `socketsecurity/core/streaming.py` — `setup_streaming()` wires the
  uploader and a terminal handler into the `socketcli` and
  `socketdev` loggers, forces both loggers to `DEBUG` so the upload
  captures the full history regardless of `--enable-debug` state,
  and returns a teardown callable for the caller to register with
  `atexit`. Module-level setters `set_run_status()` and
  `set_report_run_id()` propagate the terminal status and the
  associated full-scan id into `finalize_cli_run`.

### Existing files
- `socketsecurity/config.py` — adds `--upload-logs` /
  `--upload_logs` opt-in flag (advanced group, default off). When
  set, the CLI sends `share_logs=true` in the register payload.
- `socketsecurity/socketcli.py` — always calls `setup_streaming()`
  after `CliClient` init (register is cheap and the server may
  force-enable for an org); registers the teardown via `atexit` only
  when the server returned a run_id. Captures `diff.id` at a single
  chokepoint after the diff-producing branches converge and threads
  it through `set_report_run_id()`, guarded against the
  `NO_DIFF_RAN` / `NO_SCAN_RAN` sentinel values. Exception handlers
  in `cli()` now call `set_run_status(...)`:
  - `KeyboardInterrupt` → `cancelled`
  - any `SystemExit` with non-zero code → `failure`
  - uncaught `Exception` → `failure`
  - clean exit → default `success`

  The `SystemExit` handling is load-bearing: several failure paths
  in `main_code()` call `sys.exit(3)` directly, which bypasses
  `except Exception` (since `SystemExit` is a `BaseException`
  subclass, not `Exception`).

## Test plan

- [x] `uv run pytest tests/unit` — 171 unit tests pass, including
  the new tests in `test_cli_run.py`, `test_log_uploader.py`, and
  `test_streaming.py` (covering opt-in/opt-out register paths and
  the disabled-by-server response shape).
- [x] End-to-end against a local Socket backend with the matching
  endpoints applied:
  - `--upload-logs` set, server says enabled → full CLI log captured,
    run reports `status: success`, `report_run_id` populated with
    the full-scan id and joinable to the underlying `full_scans` row
  - `--upload-logs` set, simulated `RuntimeError` → run reports
    `status: failure`, log captured up to the crash
  - 15s continuous-log run → batches arrive at the backend at ~5s
    intervals while the CLI is still running
- [x] Without `--upload-logs` (and no org-side override) → CLI calls
  register, gets `log_streaming_enabled: false`, sends no further
  streaming requests; scan otherwise unchanged
- [x] Register failure (404 / 401 / network down) → CLI silently
  degrades, scan continues normally, no exception surfaces

## Public Changelog

A new opt-in `--upload-logs` flag uploads the CLI's log output to the Socket backend for the duration of the run, alongside a per-run status (`in_progress` / `success` / `failure` / `cancelled`). The transcript is captured regardless of the local `--enable-debug` state; the existing terminal verbosity is unchanged. The Socket backend can also force-enable streaming for specific orgs regardless of the flag. Default off; CLI runs without the flag are unaffected.


socketsecurity/config.py
```diff
@@ -139,6 +139,8 @@ class CliConfig:
     ignore_commit_files: bool = False
     disable_blocking: bool = False
     disable_ignore: bool = False
+    upload_logs: bool = False
+    decline_logs: bool = False
```
BarrensZeppelin: These two values are connected and would be more cleanly modelled by a single field with 3 possible values.

socketsecurity/core/streaming.py
BarrensZeppelin, line 1: The global state and `atexit.register` aren't that nice.

Would it be possible to refactor this module into a context manager.
The `set_run_status` and `set_report_run_id` methods could live on the manager.
The CLI's main function could then do `with setup_streaming(...) as streaming_logs: ...`

socketsecurity/core/streaming.py
```diff
@@ -0,0 +1,87 @@
...
+    saved_levels = (cli_logger.level, sdk_logger.level)
+    saved_propagate = (cli_logger.propagate, sdk_logger.propagate)
+    cli_logger.setLevel(logging.DEBUG)
+    sdk_logger.setLevel(logging.DEBUG)
+    cli_logger.propagate = False
+    sdk_logger.propagate = False
+    cli_logger.addHandler(terminal_handler)
+    sdk_logger.addHandler(terminal_handler)
+    cli_logger.addHandler(upload_handler)
+    sdk_logger.addHandler(upload_handler)
```
BarrensZeppelin: Nit: use a loop over `(cli_logger, sdk_logger)`.

socketsecurity/core/streaming.py
```diff
@@ -0,0 +1,87 @@
...
+        cli_logger.removeHandler(upload_handler)
+        sdk_logger.removeHandler(upload_handler)
+        log_uploader.stop()
+        finalize_cli_run(client, run_id, status=_run_status, report_run_id=_report_run_id)
+        cli_logger.removeHandler(terminal_handler)
+        sdk_logger.removeHandler(terminal_handler)
+        cli_logger.setLevel(saved_levels[0])
+        sdk_logger.setLevel(saved_levels[1])
+        cli_logger.propagate = saved_propagate[0]
+        sdk_logger.propagate = saved_propagate[1]
```
BarrensZeppelin: Nit: use a loop over `zip((cli_logger, sdk_logger), saved_levels, saved_propagate)`

socketsecurity/core/streaming.py
BarrensZeppelin: With this design the `set_run_status` method doesn't need to exist at all, it should be handled inside the manager's `__exit__`

socketsecurity/core/log_uploader.py
```diff
@@ -0,0 +1,122 @@
...
+log = logging.getLogger(__name__)
+
+_FLUSH_GUARD = threading.local()
+
+_LEVEL_MAP = {
+    logging.DEBUG: "DEBUG",
+    logging.INFO: "INFO",
+    logging.WARNING: "WARN",
+    logging.ERROR: "ERROR",
+    logging.CRITICAL: "ERROR",
```
BarrensZeppelin: Use `logging.getLevelName`

socketsecurity/core/log_uploader.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        self._thread.start()
+
+    def stop(self, timeout: float = 2.0) -> None:
+        if self._thread is None:
+            self._flush()
+            return
+        self._stop.set()
+        self._thread.join(timeout=timeout)
+        self._thread = None
```
BarrensZeppelin:
```suggestion
        if self._thread is not None:
       		self._stop.set()
        	self._thread.join(timeout=timeout)
        	self._thread = None
```

socketsecurity/core/log_uploader.py
```diff
@@ -0,0 +1,122 @@
...
+
+class UploadingLogHandler(logging.Handler):
+    def __init__(self, uploader: BatchedLogUploader, context: str = "socket-python-cli"):
+        super().__init__()
+        self._uploader = uploader
+        self._context = context
+
+    def emit(self, record: logging.LogRecord) -> None:
+        if getattr(_FLUSH_GUARD, "active", False):
+            return
```
BarrensZeppelin, line 103: This looks racy since the flush runs in a separate thread. Can't real logs get dropped if they're added while a flush is happening?

BarrensZeppelin (changes requested):

socketsecurity/config.py
```diff
@@ -866,6 +876,35 @@ def create_argument_parser() -> argparse.ArgumentParser:
...
+        help="Explicitly opt out of uploading CLI logs to the Socket backend, even "
+             "when an org-level override would otherwise enable it. Mutually "
+             "exclusive with --upload-logs."
+    )
+    advanced_group.add_argument(
+        "--no_upload_logs",
+        dest="decline_logs",
+        action="store_true",
+        help=argparse.SUPPRESS
+    )
```
BarrensZeppelin: These can use the same `dest` (`upload_logs`) with the `store_const` action and an appropriate `const`. They can also be made mutually exclusive at the `argparse` level with a mutually exclusive group.

socketsecurity/core/log_uploader.py
barslev: The guard is thread-local (`_FLUSH_GUARD = threading.local()`), so it only suppresses emits on the **uploader thread itself** — which is exactly the recursion case we want to block (cli_client's own debug log lines during the in-flight HTTP request would otherwise re-enqueue into the buffer being uploaded).

Emits from any other thread during a flush see `active=False` for their own thread, take the lock in `add()`, and append to the freshly-swapped-in buffer; they're picked up on the next flush. Nothing is dropped.

Added a deterministic regression test in d630013 that proves this: it parks the uploader thread inside `_flush()` via an `Event`, emits a real record from the main thread while `_FLUSH_GUARD.active` is set on the uploader thread, then asserts the record ships in the next batch. Ran it 20x locally, no flakiness — handoff is event-driven, not time-driven.

BarrensZeppelin:
I accidentally left two comments on a commit instead of the PR. I can port them when I'm at a PC again. 🙂

socketsecurity/core/streaming.py
```diff
@@ -0,0 +1,136 @@
...
+            return "success"
+        if issubclass(exc_type, KeyboardInterrupt):
+            return "cancelled"
+        # SystemExit with code 0 / None is a clean exit; non-zero codes signal failure.
+        if issubclass(exc_type, SystemExit) and not getattr(exc_val, "code", None):
+            return "success"
+        return "failure"
+
+
+def setup_streaming(
```
BarrensZeppelin: It looks like setup_streaming is just an alias for StreamingLogs, so maybe it should be dropped?

socketsecurity/socketcli.py
```diff
@@ -126,6 +127,8 @@ def cli():
             sys.exit(2)
         else:
             sys.exit(0)
+    except SystemExit:
+        raise
```
BarrensZeppelin:
```suggestion
```

socketsecurity/socketcli.py
```diff
@@ -183,644 +186,633 @@ def main_code():
...
-        org_response = sdk.org.get(use_types=True)
-        organizations = org_response.get("organizations", {})
+
+    with setup_streaming(
+        client=client,
+        cli_logger=log,
+        sdk_logger=socket_logger,
+        client_version=config.version,
+        share_logs=config.upload_logs is True,
+        decline_logs=config.upload_logs is False,
```
BarrensZeppelin: Can you move the share/decline disambiguation to the point where you interact with the API (to keep the invalid state impossible until it's unavoidable)?

socketsecurity/core/cli_run.py
```diff
@@ -0,0 +1,78 @@
...
+from .exceptions import APIFailure
+
+log = logging.getLogger("socketcli")
+
+
+def register_cli_run(
+    client: CliClient,
+    client_version: str,
+    share_logs: bool,
+    decline_logs: bool,
```
BarrensZeppelin: These should also be refactored to `upload_logs: bool | None = None`, right?

BarrensZeppelin (approved):
LGTM 🎉

PR #202, opened 2026-05-18
Add initial bun and vlt lockfile support

rreusser:
The python CLI fails to locate `bun.lock` and `vlt-lock.json` files, preventing them from being analyzed by developers using the CLI. This PR adds them to the hard-coded list.

⚠️ **Note:** Additional backend work is still required to properly ingest and scan these filetypes.

## Root Cause

Experimental bun/vlt SBOM resolution support was added but not propagated to the CLI's list of supported files.

## Fix

This PR simply adds the patterns to the hard-coded list of patterns.

## Public Changelog

Add bun.lock and vlt-lock.json to supported manifest types.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "socketsecurity"
-version = "2.2.86"
+version = "2.2.88"
 requires-python = ">= 3.11"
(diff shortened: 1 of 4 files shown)


lelia (approved):
thanks for doing this! added unit test coverage for the new lockfile patterns to get the GHA workflows unstuck - otherwise LGTM.

PR #204, opened 2026-05-18
Fix uv lockfile sync + version incrementation checks

lelia:
## Summary

- Revs `uv.lock` up to v2.2.89 to account for mandatory Version Check (was previously v2.2.88 from v2.2.86)
- Updates Python hook to sync uv lockfile is with any `pyproject.toml` and `__init__.py` version bumps
- Updates Version Check workflow to ensure the `uv.lock` file contains the incremented version


rreusser (approved):
Nice! Love this sort of check. 👍

PR #221, opened 2026-06-02
Always omit license details from full-scan diff request

lelia:
## Summary

Follow-on PR to the fixes introduced in #211. Now that SocketDev/socket-sdk-python#84 is released as `socketdev v3.1.2`, this PR raises the CLI SDK floor to `socketdev>=3.1.2` and makes the full-scan diff request (`fullscans.stream_diff`) always send `include_license_details=false`.

This prevents the large-repo truncation crash (`Unterminated string` / JSON parse failure) from recurring even when the user does not pass `--exclude-license-details`.

## Scope of `--exclude-license-details`

`--exclude-license-details` still controls the human-facing dashboard report URL (`?include_license_details=false`), but no longer affects the internal diff payload. The CLI help text and changelog now describe that narrower scope.

License artifact output is unchanged: `--generate-license` continues to fetch license details from the dedicated PURL endpoint before writing Socket/FOSSA legal artifacts.

## Versioning

This is a deliberate default-behavior change, so the CLI version is bumped to `2.4.0`.

## Tests

- `uv run --extra test pytest tests/core/test_sdk_methods.py tests/core/test_package_and_alerts.py tests/unit/test_socketcli.py tests/unit/test_fossa_compat.py -q` (`55 passed`)

Fixes: CE-224


flowstate (approved):
LGTM with one nit:

The CHANGELOG entry could explicitly call out the --exclude-license-details scope narrowing as a "soft breaking change for flag-scripted use."

lelia:
> LGTM with one nit:
> 
> The CHANGELOG entry could explicitly call out the --exclude-license-details scope narrowing as a "soft breaking change for flag-scripted use."

@flowstate fixed in: https://github.com/SocketDev/socket-python-cli/pull/221/commits/970fb554f0f0a4383ca3ce56dc09bac81827fbff

PR #230, opened 2026-06-08
Pin @coana-tech/cli version; make reachability auto-update opt-in

mtorp:
## Why

The Python CLI auto-updated the reachability (Coana) engine to the **latest** published version on every `--reach` run via `npm install -g @coana-tech/cli`. Automatically pulling a brand-new engine version — without the operator opting in — is undesirable for environments that need to review/approve dependency updates before adopting them. This makes the engine version explicit and opt-in, and aligns the launcher behavior with the Socket Node CLI.

### Pin the engine version (no silent auto-update)
- Reachability now runs a fixed `@coana-tech/cli` version baked into the release (`DEFAULT_COANA_CLI_VERSION = 15.3.24`, matching the Node CLI's pin). The engine version changes only through a standard `pip` upgrade of this CLI.
- The `npm install -g @coana-tech/cli` step (the auto-update) is removed entirely. An existing global install is left untouched — never auto-updated or downgraded.
- Opt into newest with `--reach-version latest`; pin an explicit version with `--reach-version <semver>`.

### Launcher (parity with the Socket Node CLI's `dlx` path)
- Run the engine via `npx --yes --force @coana-tech/cli@<ver> …` — the exact flags the Node CLI passes for coana. `--yes` skips npx's interactive install prompt so non-interactive/CI runs don't hang. (Note: `--force` does **not** force a re-download of an already-cached pinned version; it mirrors the Node CLI's flags rather than bypassing the cache.)
- **`npm install` + `node` fallback:** if the `npx` launcher is missing or dies *before* the engine starts (spawn error / signal / exit ≥128), fall back to `npm install --no-save --no-package-lock --no-audit --no-fund --prefix <tmpdir> @coana-tech/cli@<ver>`, resolve the `bin` from the installed `package.json`, and run via `node`. Small positive exit codes are real engine failures and are **not** retried. The resolved install is cached per version for the process lifetime (installs once).
- Tunable via `SOCKET_CLI_COANA_FORCE_NPM_INSTALL` (use the fallback as the primary path) and `SOCKET_CLI_COANA_DISABLE_NPM_FALLBACK` (never fall back). `node` is now part of the up-front prerequisite check. Strips `npm_package_*` env vars before spawning to avoid `E2BIG` in large monorepos.

### Docs / changelog
- Updated `--reach-version` help, `docs/cli-reference.md`, and `CHANGELOG.md` (2.4.7).

## Tests

`tests/unit/test_reachability.py` (29 tests): spec resolution, the `--yes --force` flags, `npm_package_*` stripping, the launcher-failure heuristic, every fallback branch (launcher failure / missing `npx` / `FORCE_NPM_INSTALL` / `DISABLE_NPM_FALLBACK`), per-version install caching, the `node`-missing error, and real `_resolve_coana_bin` / `_build_coana_node_cmd` parsing. Full unit suite: 282 passed, 2 (pre-existing, unrelated) skips.

## Review

A comprehensive automated review was run on this PR; its Major findings (accurate npx-cache wording, a per-version fallback cache to avoid temp-dir leak / re-install, and a clear error + prereq for a missing `node`) have been addressed in the latest commit.

## Maintenance note

`DEFAULT_COANA_CLI_VERSION` is a deliberate, reviewable bump per release rather than an implicit "always latest". Reviewers should expect to update it (and the `currently 15.3.24` references in the docs/CHANGELOG) together.


docs/cli-reference.md
```diff
@@ -240,7 +240,7 @@ If you don't want to provide the Socket API Token every time then you can use th
 | Parameter                        | Required | Default | Description                                                                                                                |
 |:---------------------------------|:---------|:--------|:---------------------------------------------------------------------------------------------------------------------------|
 | `--reach`                          | False    | False   | Enable reachability analysis to identify which vulnerable functions are actually called by your code. Creates a tier-1 full-application reachability scan (`scan_type=socket_tier1`). |
-| `--reach-version`                  | False    | latest  | Version of @coana-tech/cli to use for analysis                                                                             |
+| `--reach-version`                  | False    | *pinned* | Version of @coana-tech/cli to use. Defaults to the version pinned to this CLI release (currently `15.3.24`), so the engine only changes when you upgrade the Socket CLI. Pass `latest` to always use the newest published version (opt-in auto-update), or an explicit version (e.g. `1.2.3`) to pin it. |
```
BarrensZeppelin: What's the point of the `Default` column if it doesn't contain the default value?

docs/cli-reference.md
```diff
@@ -313,7 +314,11 @@ Sample config files:
...
+The CLI runs a pinned `@coana-tech/cli` version via `npx --yes --force` (the same flags the Socket Node CLI passes for coana); it does **not** auto-update the engine or install it globally. `--yes` skips npx's interactive install prompt so non-interactive/CI runs don't hang. If the `npx` launcher is unavailable or fails before the engine starts, the CLI falls back to `npm install`-ing the pinned version into a temp directory and running it via `node`. Pass `--reach-version latest` to opt into the newest published version. Use `--reach` to enable reachability analysis during a full scan, or add `--only-facts-file` (with `--reach`) to submit only the reachability facts file (`.socket.facts.json`) when creating the full scan.
+
+The launcher fallback can be tuned via environment variables:
+- `SOCKET_CLI_COANA_FORCE_NPM_INSTALL` — skip `npx` entirely and always use the `npm install` + `node` path (useful where `npx` is known-broken).
+- `SOCKET_CLI_COANA_DISABLE_NPM_FALLBACK` — never fall back; surface the `npx` failure directly.
```
BarrensZeppelin, line 321: These env variables seem related, so maybe they should be merged into one with two possible values?

socketsecurity/core/tools/reachability.py
```diff
@@ -7,11 +7,22 @@
...
+import tempfile
 
 from socketsecurity import __version__
 
 log = logging.getLogger(__name__)
 
+# Pinned @coana-tech/cli version. Bumped deliberately per Python CLI release so the
+# reachability engine version only changes through a standard pip upgrade (advance notice).
+# Pass --reach-version latest to opt into the newest published version instead.
+DEFAULT_COANA_CLI_VERSION = "15.3.24"
```
BarrensZeppelin:
```suggestion
DEFAULT_COANA_CLI_VERSION: Final = "15.3.24"
```

Requires updating the `from typing import ...` import above.

socketsecurity/core/tools/reachability.py
```diff
@@ -31,70 +42,31 @@ def __init__(self, sdk: socketdev, api_token: str):
...
     
-    def _ensure_coana_cli_installed(self, version: Optional[str] = None) -> str:
+    def _resolve_coana_package_spec(self, version: Optional[str] = None) -> str:
         """
-        Check if @coana-tech/cli is installed, and install/update it if needed.
-        
+        Resolve the @coana-tech/cli package spec to run with npx.
+
+        We pass an exact, versioned spec to npx so it runs a deterministic version from its
+        own cache (fetching once if absent). We intentionally do NOT ``npm install -g`` here:
```
BarrensZeppelin: Why would we run `npm install -g` here in the first place?

socketsecurity/core/tools/reachability.py
```diff
@@ -303,7 +269,173 @@ def run_reachability_analysis(
...
+        """
+        effective_version = self._resolve_coana_version(version)
+        coana_env = self._sanitize_coana_env(env)
+        disable_fallback = bool(os.environ.get("SOCKET_CLI_COANA_DISABLE_NPM_FALLBACK"))
+
+        if os.environ.get("SOCKET_CLI_COANA_FORCE_NPM_INSTALL"):
+            return self._spawn_coana_via_npm_install(coana_args, effective_version, coana_env, cwd)
+
+        package_spec = f"@coana-tech/cli@{effective_version}"
+        # --yes skips npx's install confirmation; --force matches the Node CLI's coana flags.
```
BarrensZeppelin: This is duplicated in the docstring.

socketsecurity/core/tools/reachability.py
```diff
@@ -303,7 +269,173 @@ def run_reachability_analysis(
...
+        cwd: str,
+    ) -> int:
+        """Run coana for the given args, returning the process exit code.
+
+        Primary path: ``npx --yes --force @coana-tech/cli@<version> ...`` — the exact flags
+        the Socket Node CLI passes for coana. ``--yes`` skips npx's interactive install
+        confirmation so non-interactive/CI runs don't hang. ``--force`` matches the Node CLI
+        (it opts out of npm's prompts/protections and refreshes within-range specs); note it
+        does NOT force a re-download of an already-cached pinned version — npx still reuses a
+        cached pinned package, so this is parity with the Node CLI, not a cache bypass.
```
BarrensZeppelin: Maybe 4 lines is too much to explain what `--force` does and doesn't do.

tests/unit/test_reachability.py
```diff
@@ -21,19 +24,32 @@ def analyzer():
...
-def _run(analyzer, mocker, **kwargs):
-    """Invoke run_reachability_analysis with npm/npx/coana mocked; return (cmd, env)."""
-    mocker.patch.object(analyzer, "_ensure_coana_cli_installed", return_value="@coana-tech/cli")
+@pytest.fixture(autouse=True)
+def _clear_coana_install_cache():
+    """The npm-install fallback caches resolved script paths in a module-level dict; isolate tests."""
+    reachability._INSTALLED_COANA_SCRIPT_PATHS.clear()
+    yield
+    reachability._INSTALLED_COANA_SCRIPT_PATHS.clear()
+
```
BarrensZeppelin, line 35:
```suggestion
    reachability._INSTALLED_COANA_SCRIPT_PATHS.clear()
    try:
    	yield
    finally:
    	reachability._INSTALLED_COANA_SCRIPT_PATHS.clear()
```

tests/unit/test_reachability.py
```diff
@@ -104,3 +120,265 @@ def test_repo_branch_env_absent_when_none(analyzer, mocker):
...
+def test_resolve_spec_latest_opt_in(analyzer):
+    """'latest' opts into the newest published version."""
+    assert analyzer._resolve_coana_package_spec("latest") == "@coana-tech/cli@latest"
+
+
+def test_resolve_spec_is_always_versioned(analyzer):
+    """Never the bare '@coana-tech/cli' (which would let npx pick a stray global version)."""
+    for version in (None, "latest", "1.2.3", " 1.2.3 "):
+        assert analyzer._resolve_coana_package_spec(version).startswith("@coana-tech/cli@")
+
```
BarrensZeppelin: Use `pytest.mark.parametrize`

tests/unit/test_reachability.py
```diff
@@ -104,3 +120,265 @@ def test_repo_branch_env_absent_when_none(analyzer, mocker):
...
+
+def test_npx_runs_explicit_version(analyzer, mocker):
+    cmd, _ = _run(analyzer, mocker, version="9.9.9")
+    assert _spec_in(cmd) == "@coana-tech/cli@9.9.9"
+
+
+def test_npx_runs_latest_when_opted_in(analyzer, mocker):
+    cmd, _ = _run(analyzer, mocker, version="latest")
+    assert _spec_in(cmd) == "@coana-tech/cli@latest"
```
BarrensZeppelin: Same

tests/unit/test_reachability.py
```diff
@@ -104,3 +120,265 @@ def test_repo_branch_env_absent_when_none(analyzer, mocker):
...
+    monkeypatch.setenv("npm_config_registry", "https://example.test")
+    _, env = _run(analyzer, mocker)
+    assert "npm_package_dependencies_foo" not in env
+    assert env.get("npm_config_registry") == "https://example.test"
+
+
+# --- npm-install + node fallback (when the npx launcher fails before coana starts) ---
+
+
+def test_launcher_failure_heuristic():
```
BarrensZeppelin: Same

socketsecurity/core/tools/reachability.py
```diff
@@ -7,11 +7,22 @@
...
 
+# Pinned @coana-tech/cli version. Bumped deliberately per Python CLI release so the
+# reachability engine version only changes through a standard pip upgrade (advance notice).
+# Pass --reach-version latest to opt into the newest published version instead.
+DEFAULT_COANA_CLI_VERSION = "15.3.24"
+
+# Resolved @coana-tech/cli script paths from the npm-install fallback, keyed by version.
+# Lives for the process lifetime so repeated fallback invocations install only once
+# (mirrors the Node CLI's installedCoanaScriptPathsByVersion).
+_INSTALLED_COANA_SCRIPT_PATHS: Dict[str, str] = {}
```
BarrensZeppelin, line 26: Do we want to clean these up afterwards? Python has `atexit.register`

BarrensZeppelin (approved):
LGTM ✅

socketsecurity/core/tools/reachability.py
mtorp: I guess it comes from how the previous implementation behaved.

docs/cli-reference.md
mtorp: We can insert the default value. We just have to remember to keep it up-to-date then.

BarrensZeppelin: We already have to do that, right? (The current values are in the help text)

socketsecurity/core/tools/reachability.py
BarrensZeppelin: Maybe it belongs in the `_spawn_coana` function?

docs/cli-reference.md
mtorp: The Socket Node CLI uses two separate env vars here too — `SOCKET_CLI_COANA_FORCE_NPM_INSTALL` and `SOCKET_CLI_COANA_DISABLE_NPM_FALLBACK` (see `src/utils/dlx.mts`). They're orthogonal rather than two values of one setting: one forces the npm-install path as the primary launcher (skip npx), the other forbids the npm-install fallback (npx only) — and the default is neither (npx with fallback). Keeping them as two separate flags keeps the Python and Node CLIs consistent, so I've left them as-is.

mtorp: Lol, I specifically asked Claude not to reply here 🤷

mtorp: I'll make it a followup item to change it for both the Node and the Python CLI. I don't think anybody are using these vars yet, so should be fine.

docs/cli-reference.md
```diff
@@ -240,7 +240,7 @@ If you don't want to provide the Socket API Token every time then you can use th
...
+| `--reach-version`                  | False    | 15.3.24 | Version of @coana-tech/cli to use. Defaults to the pinned version that ships with this CLI release, so the engine only changes when you upgrade the Socket CLI. Pass `latest` to always use the newest published version (opt-in auto-update), or an explicit version (e.g. `1.2.3`) to pin it. |
 | `--reach-analysis-timeout`         | False    | *coana* | Timeout in seconds for the reachability analysis. Omitted by default, so coana applies its own (currently 600s). Alias: `--reach-timeout` |
 | `--reach-analysis-memory-limit`    | False    | *coana* | Memory limit in MB for the reachability analysis. Omitted by default, so coana applies its own (currently 8192). Alias: `--reach-memory-limit` |
 | `--reach-concurrency`              | False    | *coana* | Control parallel analysis execution (must be >= 1). Omitted by default, so coana applies its own (currently 1)             |
```
BarrensZeppelin: These have the same issue.

docs/cli-reference.md
BarrensZeppelin: I guess `info` is technically the default?

docs/cli-reference.md
mtorp: Good catch 👍

PR #232, opened 2026-06-10
Retry transient full-scan upload failures (502/503/504/408, dropped connections)

mtorp:
## Problem

The full-scan upload (`POST /v0/orgs/<org>/full-scans`) makes exactly one attempt. When it fails, the entire run dies — for `--reach` runs that means a completed multi-hour tier-1 reachability analysis produces no scan and no report.

Uploads can fail transiently at the gateway/connection level — an HTTP 502/503/504/408, a dropped or reset connection, or a client-side timeout — without the server having created the scan. A retried upload almost always succeeds.

## Change

`Core.create_full_scan` now retries the `fullscans.post` call:

- **Up to 3 total attempts**, waiting ~10s before attempt 2 and ~30s before attempt 3 (plus up to 2s jitter so CI fleets don't retry in lock-step), with a `log.warning` before each retry.
- **Retries transient failures only**, as classified by the SDK: requires `socketdev>=3.3.0` (SocketDev/socket-sdk-python#93), where `do_request` records the HTTP status code on every exception it raises and `APIFailure.is_transient_error()` owns the classification — true for HTTP 408/502/503/504, dropped/reset connections, and client-side timeouts. The CLI encodes no knowledge of the SDK exception hierarchy and parses no message text, so SDK restructuring can't silently break the classification.
- **Never retried**: deterministic errors (400/401/403/404/429), responses that came back with an error payload, and unexpected errors wrapped in a bare `APIFailure`.

> **Note**: `uv.lock` is regenerated once `socketdev 3.3.0` is released to PyPI (blocked on SocketDev/socket-sdk-python#93); CI is expected to fail on dependency resolution until then.

### Safety

- **No duplicate scans**: in these failure modes the server never finished reading the request body, so no scan was created — a retry cannot duplicate one. In the rare case where a gateway timeout races a request the server later completes, the extra scan is benign and superseded by the retry (as if the CLI had run twice).
- **Files are rebuilt per attempt**: `fullscans.post(use_lazy_loading=True)` rebuilds its `LazyFileLoader`s from the plain paths on every call, so re-invoking it is safe (the loaders from a failed attempt are exhausted/closed and must not be reused).
- **Temp `.br` facts files survive retries**: the retry loop lives inside the existing `try`/`finally`, so the brotli-compressed `.socket.facts.json.br` siblings are only cleaned up after the last attempt.

## Testing

New `tests/unit/test_full_scan_retry.py` (16 tests): retry on 502 then success, exhausting attempts re-raises, retry on catch-all 408/503/504, retry on connection-level errors, no retry on 400 / dedicated 4xx classes / error payloads, `.br` temp-file lifecycle across retries (success and exhaustion), and that the retry decision delegates to `APIFailure.is_transient_error()` (the classification truth table itself is tested in the SDK, next to the code that raises). Full suite against socketdev 3.3.0: **364 passed, 2 pre-existing skips**.

### DIFF
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -84,11 +84,10 @@ TIER1_FINALIZE_BACKOFF_SECONDS = 1.0
 
 # Full scan upload retry policy. An upload can fail transiently at the gateway/connection
 # level (an HTTP 502/503/504/408, a dropped or reset connection, or a client-side timeout)
-# without the server having created the scan; a retried upload almost always succeeds.
-# In these failure modes no scan was created, so a retry does not duplicate one. (A
-# duplicate is possible only if a gateway timeout races a request the server later
-# completes; that is benign - the retried scan supersedes the orphaned one, same as
-# running the CLI twice.)
+# without the server having created the scan. In these failure modes no scan was created,
+# so a retry does not duplicate one. (A duplicate is possible only if a gateway timeout
+# races a request the server later completes; that is benign - the retried scan supersedes
+# the orphaned one, same as running the CLI twice.)
 FULL_SCAN_UPLOAD_MAX_ATTEMPTS = 3
 # Wait before retry attempt 2 and attempt 3 respectively (plus a little jitter so a fleet of
 # CI jobs hitting the same failure doesn't retry in lock-step).
@@ -836,8 +835,7 @@ class Core:
         upload_files, compressed_temp_files = self._compress_facts_files_for_upload(files)
         try:
             # Retry transient gateway/timeout failures (502/503/504/408, dropped connections,
-            # timeouts) with increasing waits; such failures are intermittent and a retried
-            # upload almost always succeeds. In these failure modes the server never finished
+            # timeouts) with increasing waits. In these failure modes the server never finished
             # reading the request body, so no scan was created and a retry does not duplicate
             # one (see the retry-policy comment above FULL_SCAN_UPLOAD_MAX_ATTEMPTS). fullscans.post()
             # rebuilds its lazy file loaders from the plain paths in upload_files on every call,
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import random
 import re
 import sys
 import tarfile
@@ -13,7 +14,12 @@ from typing import Dict, List, Tuple, Set, TYPE_CHECKING, Optional
 if TYPE_CHECKING:
     from socketsecurity.config import CliConfig
 from socketdev import socketdev
-from socketdev.exceptions import APIFailure
+from socketdev.exceptions import (
+    APIBadGateway,
+    APIConnectionError,
+    APIFailure,
+    APITimeout,
+)
 from socketdev.fullscans import FullScanParams, SocketArtifact
 from socketdev.org import Organization
 from socketdev.repos import RepositoryInfo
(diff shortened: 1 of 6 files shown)


socketsecurity/core/__init__.py
```diff
@@ -76,6 +82,48 @@
 TIER1_FINALIZE_MAX_ATTEMPTS = 3
 TIER1_FINALIZE_BACKOFF_SECONDS = 1.0
 
+# Full scan upload retry policy. An upload can fail transiently at the gateway/connection
+# level (an HTTP 502/503/504/408, a dropped or reset connection, or a client-side timeout)
+# without the server having created the scan; a retried upload almost always succeeds.
+# In these failure modes no scan was created, so a retry does not duplicate one. (A
+# duplicate is possible only if a gateway timeout races a request the server later
+# completes; that is benign - the retried scan supersedes the orphaned one, same as
+# running the CLI twice.)
```
mtorp: Remove comment about a retry almost always succeeds.

socketsecurity/core/__init__.py
```diff
@@ -76,6 +82,47 @@
...
+# Transient gateway/timeout HTTP statuses that the SDK does NOT raise as a dedicated
+# exception class (502 has APIBadGateway; 408/503/504 surface as the catch-all APIFailure
+# with the status only present in the message text - see _is_transient_full_scan_upload_error).
+FULL_SCAN_UPLOAD_RETRYABLE_STATUS_CODES = frozenset({408, 503, 504})
+# Matches the status code the SDK embeds in catch-all APIFailure messages
+# (socketdev/core/api.py: "Bad Request: HTTP original_status_code:<code> ...").
+_API_FAILURE_STATUS_CODE_RE = re.compile(r"original_status_code:(\d{3})")
+
+
+def _is_transient_full_scan_upload_error(error: Exception) -> bool:
```
BarrensZeppelin: This design is vulnerable to restructuring in the `socketdev` dependency.
If a new exception subclass is introduced this code breaks (silently).

Maybe we could either:
1. Move the functionality to `socketdev` (put something like `is_transient_error()` on APIFailure.
2. Put safeguards in place here to fail loudly if invariants around the classes/status codes are broken.

socketsecurity/core/__init__.py
```diff
@@ -787,7 +834,33 @@ def create_full_scan(self, files: List[str], params: FullScanParams, base_paths:
...
+                    res = self.sdk.fullscans.post(upload_files, params, use_types=True, use_lazy_loading=True, max_open_files=50, base_paths=base_paths)
+                    break
+                except APIFailure as error:
+                    if attempt >= FULL_SCAN_UPLOAD_MAX_ATTEMPTS or not _is_transient_full_scan_upload_error(error):
+                        raise
+                    backoff_index = min(attempt, len(FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS)) - 1
+                    wait_seconds = FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS[backoff_index] + random.uniform(
+                        0, FULL_SCAN_UPLOAD_BACKOFF_JITTER_SECONDS
+                    )
+                    # SDK error messages can span many lines (path + response headers); the
```
BarrensZeppelin: I think this part can be made cleaner by:
1. change `FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS` to `(10., 30., None)`
2. do: `for attempt, backoff_seconds in enumerate(FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS, start=1): ...`
3. change the `attempt >= ...` condition to `backoff_seconds is None`
4. remove `FULL_SCAN_UPLOAD_MAX_ATTEMPTS` or compute as `len(FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS)`

BarrensZeppelin (approved):
LGTM

socketsecurity/core/__init__.py
mtorp: Good point. I'll try to move the logic into the SDK.

mtorp: Moved into the SDK in SocketDev/socket-sdk-python#93: `do_request` now records the HTTP status code on every exception it raises, and `APIFailure.is_transient_error()` owns the classification. It's status-based rather than class-based, so if a transient status (e.g. 503) ever gains a dedicated subclass it keeps working unchanged. The CLI side (487557a) now just calls `error.is_transient_error()` — no hierarchy knowledge, no message parsing.

Requires `socketdev>=3.3.0`, so CI here stays red until the SDK PR is merged and released; uv.lock gets regenerated then.

socketsecurity/core/__init__.py
mtorp: Done in 487557a: the schedule is now `(10.0, 30.0, None)`, the loop is `for attempt, backoff_seconds in enumerate(FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS, start=1)`, and the exhaustion check is `backoff_seconds is None`. `FULL_SCAN_UPLOAD_MAX_ATTEMPTS` is kept but computed as `len(FULL_SCAN_UPLOAD_BACKOFF_SCHEDULE_SECONDS)` (it's still used in the retry log message and the tests).

PR #235, opened 2026-06-12
Add `--include-dirs` flag to scan normally-excluded directories

lelia:
## Summary

The CLI excludes a fixed set of directory names from manifest discovery (`build`, `dist`, `node_modules`, `.venv`, `__pycache__`, etc.). These are sane defaults, but some projects keep manifest files under those names (e.g. `build/requirements.txt`), so there was no way to scan them. This adds a generic way to opt specific directories back in.

## Changes

- **New `--include-dirs` flag** (comma-separated directory names) — re-includes any of the default-excluded directory names in manifest discovery, e.g. `--include-dirs build,dist`. Names are matched against any path segment, mirroring how the default exclude list is applied.
- **`--include-module-folders` now works as documented** — it re-includes the JS/TS module folders (`node_modules`, `bower_components`, `jspm_packages`) as a group. It was previously accepted but had no effect.

The re-inclusion builds a new exclude set rather than mutating the shared module-level default, and is applied right after `Core` is constructed so it covers every manifest scan (including the sub-path pre-check).

## Tests

New `tests/unit/test_include_dirs.py`: config parsing (CLI + config file), `find_files` excludes `build` by default vs. includes it when un-excluded, and a guard that the shared default set is never mutated. Existing config/CLI suites pass.

Fixes CE-228.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## 2.4.10
+
+### Added: opt directories back into manifest discovery via `--include-dirs`
+
+- New `--include-dirs` flag (comma-separated directory names) that re-includes directories
+  the CLI excludes from manifest discovery by default. The default exclude list
+  (`node_modules`, `bower_components`, `jspm_packages`, `__pycache__`, `.venv`, `venv`,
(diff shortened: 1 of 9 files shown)


flowstate (approved):
lgtm!

PR #301, opened 2026-08-13
Reduce scan startup time in large repositories

lelia:
### Summary

Reduce Python CLI scan setup time in large repositories by replacing repeated filesystem traversal and broad Git fetching with single-pass, local-first implementations. Add native Buildkite pull-request context and enough diff-scan telemetry to distinguish client work from backend comparison time.

In an affected Buildkite pipeline, the preview reduced the two local bottlenecks from roughly 8 minutes combined to about 3 seconds:

| Phase               | Before    | Preview |
|---------------------|----------:|--------:|
| Git initialization  | ~4m 10s   | 1.65s   |
| Manifest discovery  | ~4m 06s   | 1.30s   |

The preview found the same 765 manifests after visiting 175,045 files. The remaining end-to-end time was dominated by backend scan comparison and is being investigated separately.

### Changes

- Discover supported manifests with one streaming `os.walk()` per scan root, pruning excluded directories and `.git` before descent. Cache successfully loaded patterns and reuse discovery results for `--sub-path` scans.
- Remove unconditional `git fetch --all`. Prefer refs already in the checkout and fetch only a missing base or head ref when a comparison requires it.
- Recognize Buildkite commit, branch, and pull-request variables directly, including GitHub SCM comment context, without GitHub Actions variable shims.
- Add INFO-level timings and counters for CLI initialization, Git setup, changed-file detection, pattern loading, manifest discovery, and diff polling.
- Log the diff-scan ID and report URL, cap polling intervals at 10 seconds, and document the API token scopes required by the diff-scan flow.
- Omit unchanged comparison artifacts when no enabled output consumes them, avoiding a large response and object graph on ordinary PR runs. Preserve the full result for strict blocking, GitLab security output, license generation, and FOSSA output.
- Make opt-in Docker previews build from the downloaded wheel and publish both amd64 and arm64 images.

### Behavior Notes

- Scan routing is unchanged: the CLI still creates a full report when no supported manifest appears in the changed-file set.
- `.git` is intentionally excluded from manifest discovery.
- Directory exclusions now honor globs, so existing defaults such as `*.egg-info` are applied as intended.
- The unchanged-artifact optimization reduces response size and client memory; it does not skip the underlying backend comparison.

### Validation

- Validated a Docker preview in the affected Buildkite pipeline: Git setup and manifest discovery fell from ~8m 16s to ~3s combined with equivalent manifest results.
- Matching-parity coverage spans every built-in manifest pattern plus casing, nested patterns, hidden directories, exclusions, symlinks, multiple roots, sorting, and deduplication.
- Git coverage includes local-ref preference, targeted-fetch fallback, Buildkite, GitHub Actions, GitLab CI, Bitbucket Pipelines, detached HEAD, and non-PR execution.
- Memory regression coverage verifies discovery remains streaming and bounded by directory width and the result set rather than repository size.
- Live API validation confirms `omit_unchanged` removes unused artifacts while preserving them for every output mode that consumes them.
- `476 passed, 2 skipped`; Ruff clean on all changed Python files; source and wheel builds verified.

Ref: CE-379

---

> [!NOTE]
> **Medium Risk**
> Changes sit on the critical path for manifest discovery, Git PR baselines, and diff-scan payloads; incorrect parity or gating could mis-route scans or strip data from strict-blocking/license outputs, though broad unit tests and streaming fallbacks limit blast radius.
> 
> **Overview**
> **Release 2.6.6** cuts local scan startup on large repos by replacing repeated per-pattern `rglob` manifest discovery with **one pruned `os.walk` per scan root** (cached API patterns, basename prefilter, directory globs like `*.egg-info`), and by **reusing** manifest lists from `--sub-path` pre-checks during scan creation.
> 
> **Git** no longer runs `git fetch --all` at init. PR/MR changed-file detection uses **local refs first** and **targeted fetches** only when a base or history is missing. **Buildkite** commit, branch, and PR variables are read natively (including default-branch vs PR routing), and **`--scm github` on Buildkite** can derive GitHub comment context from `BUILDKITE_*` when GitHub Actions vars are absent.
> 
> **Diff comparisons** always load artifacts through the cached diff-scan GET (eager create/list payloads cannot skip it). When no enabled output needs unchanged packages, polls request **`omit_unchanged=true`**; polling max interval is **10s** (down from 30s), with INFO logs for diff-scan id, report URL, poll counts, and stage timings across init, Git, discovery, and registration.
> 
> **CI/docs:** PR Docker previews build from the workspace context with **linux/amd64 and linux/arm64** (QEMU). Troubleshooting and Buildkite docs cover diff-scan token scopes and checkout expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b812a18c1e8e8610080c9b9ab85a9474b3ba95bb. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lelia:
bugbot run

lelia:
bugbot run

socketsecurity/core/__init__.py
```diff
@@ -106,8 +103,15 @@
 # minutes to compute. The timeout is a backstop against a diff scan that never
 # completes; on expiry (or any other failure of this flow) the caller falls back to the
 # legacy streaming comparison rather than failing the scan outright.
+#
+# The max interval is also the upper bound on how long a finished comparison sits
+# unnoticed between polls, which is dead time added to every PR job. Callers commonly
+# run this inside a CI step with a per-step time budget of a few minutes, so the cap is
+# kept small: a multi-minute comparison costs roughly 2x the polls of a 30s cap while
+# cutting the worst-case dead time from 30s to 10s. Diff scans log their poll count and
+# last interval on completion so this tradeoff can be re-evaluated against real timings.
```
BarrensZeppelin:
```suggestion
```
`git blame` will tell you why 10s was chosen. 🤷

socketsecurity/core/__init__.py
```diff
@@ -421,6 +455,112 @@ def format_bytes(bytes_value):
...
+
+        # PurePath.match compares pattern segments right to left, so a path-shaped glob
+        # can only match a file whose basename matches the glob's final segment. Folding
+        # those final segments into the basename prefilter lets the walk skip the path
+        # match for everything else. An empty final segment (a trailing "/") constrains
+        # nothing, so it becomes "*" and the prefilter admits every name.
+        candidate_basenames = set(literal_basenames)
+        candidate_basename_globs = set(basename_globs)
+        for pattern in path_globs:
+            final_segment = pattern.rstrip("/").rsplit("/", 1)[-1] or "*"
```
BarrensZeppelin: The only way `final_segment` becomes `"*"` is if the pattern is empty or has a suffix of `"//"`. Is that intended?

CHANGELOG.md
BarrensZeppelin, line 1: IMO this can be trimmed a bit. Claude likes to include a lot of details from the problem investigation, which is maybe not necessary for a user-facing changelog.

BarrensZeppelin (approved):
Perf improvements LGTM. ✅ 
I don't have context on Buildkite/GHA.

socketsecurity/core/__init__.py
lelia: fair point!

CHANGELOG.md
lelia: agreed! trimmed.

socketsecurity/core/__init__.py
lelia: good catch, fixed!

lelia:
bugbot run

PR #302, opened 2026-08-13
Preserve pull request context, full-scan SCM branch pipelines, and gate ignore commands

lelia:
## Summary

Behavior changes that existing pipelines will notice:

- With `--scm github` or `--scm gitlab`, only pull request and merge request events create diff scans. Every other pipeline, including default-branch pushes, creates a full scan, and `--enable-diff` no longer opts a branch run into comparison mode.
- Those branch runs no longer return a blocking exit code. A full scan has no baseline, so it cannot distinguish newly introduced alerts from pre-existing ones; the CLI behaves as if `--disable-blocking` was supplied, matching how it already treats a run with no supported manifest files. Pull request and merge request pipelines still block.
- `@SocketSecurity ignore` and `ignore-all` are honored only from a commenter with write access to the repository. On GitHub the CLI queries and caches each commenter's effective repository permission, requiring write, maintain, or admin access instead of trusting relationship labels such as `MEMBER` or `COLLABORATOR`; a 404 is a definitive denial for a non-collaborator. On GitLab it reads project membership. If either provider has a genuinely indeterminate lookup failure, the command is honored with a warning under the default `enforce` policy; `strict` rejects indeterminate access and `off` performs no check.
- `--pr-number` is auto-detected in GitHub Actions, GitLab CI and Azure Pipelines when omitted. An explicit value still wins, including `0` to disable association, and a non-numeric value such as Buildkite's `false` is normalized to `0`.
- `--scm github` and `--scm gitlab` imply the matching `--integration` unless `--integration` is supplied explicitly.

Fixes:

- Per-alert ignore commands did not work. The two matching paths disagreed on the package identifier — the comment flow matched the bare name from the alert marker, the push flow matched the ecosystem-prefixed name — so no single string worked in both, and a developer following the printed instructions saw the alert return on the next commit. Both forms are now accepted on both paths.
- Push-flow ignore telemetry used a narrower package matcher than alert suppression, so a legacy bare-name command suppressed its alert without producing a telemetry event. Telemetry now uses the same canonical matcher as suppression.
- GitHub's collaborator-permission endpoint returns 404 for a non-collaborator, but the ignore gate treated every request failure as indeterminate and honored the command under the default policy. A 404 is now cached as an unauthorized result; only genuinely unreadable lookups use the configured fallback.
- Scoped package names could not be ignored at all, because the specifier was split on every `@`. The resulting error was caught and the command discarded silently, while the developer still got the acknowledgement reaction.
- An npm scope was mistaken for an ecosystem prefix when matching, so `ignore @types/node@*` also suppressed alerts for a package named `node`.
- The legacy comment table parser unpacked each row through four consecutive splits with no bounds checks. A row it could not read raised and ended the run before status reporting, and a scoped package name in Socket's own table was enough to trigger it. Rows are now parsed defensively and an unreadable row keeps its alert reported.
- `--generate-license` and `--legal-format fossa` produced an empty attribution file on any path without a baseline, because the package list was only fetched for alert-bearing output formats.
- Namespaced packages were sent to the PURL license endpoint without their namespace, and namespaced responses could not be matched back to the scanned package. Scoped npm and namespaced Maven packages now receive their license details.
- Console-only full scans skipped the SBOM alert fetch but rendered the empty local alert list as `No issues found`. They now state that findings were not fetched for console output and link to the Socket report.
- Dependency overviews presented updated packages as new ones; added, updated, removed and replaced are now distinct.
- Diff scans lost their change link when a before/after pair had already been compared.
- `GITHUB_SERVER_URL` and `CI_SERVER_URL` were composed into a diff scan's `external_href` without the scheme and host validation applied to the sibling repository URLs read from the same environment.
- Manifest paths, sources and alert text are now escaped when they are rendered into pull request and merge request comments. None of it is markup the CLI authored — the paths come from the scanned repository, the text from the API — so each is escaped at the point it is interpolated. Slack, Jira and console output are unchanged, since none of them render HTML.
- Shared security comment copy described GitLab merge request output as Socket for GitHub.
- The GitLab authentication fallback never ran. When a token's type cannot be inferred from its shape the CLI guesses between Bearer and PRIVATE-TOKEN and retries once under the other scheme on a 401, but the retry caught an exception type the HTTP client never raises, the HTTP status was discarded before it could be checked, and the CLI's `APIFailure` was an independent type from the SDK exception of the same name that every handler catches. A misclassified token failed the run instead of falling back.

Added:

- Console findings and GitHub/GitLab security comments show the patched version when the API supplies `firstPatchedVersionIdentifier`.

## Implementation notes

Pull request context is resolved from explicit configuration first, then from standard provider CI variables, and the normalized number is stored before any comment adapter reads it.

Repository slugs and hosts are recovered from `BUILDKITE_REPO` through one shared parser, so nested GitLab subgroups and self-hosted installations produce correct links without extra configuration.

The diff scan link is sent with `on_duplicate=update`. Deployments that answer 409 regardless keep the existing conflict-resolution fallback.

Ignore authorization is gated where comments are bucketed rather than at each consumer, so a rejected command reaches neither the alert filter nor the ignore telemetry, and no acknowledgement reaction is added to a comment that was not honored. GitHub effective permissions are cached per commenter, including negative 404 results; GitLab membership is cached per run. Generated ignore commands are ecosystem-qualified; parsing stays compatible with older bare-name replies and splits a scoped name at its final `@`. Telemetry delegates to that same matcher.

Full scans carry explicit output metadata so console summaries can distinguish “no fetched findings” from a verified empty result without paying for an otherwise unnecessary SBOM fetch.

## Testing

- `688 passed, 0 skipped` across `tests/unit` and `tests/core`
- `ruff check` clean across the repository
- `uv lock --locked` succeeds
- Source distribution and wheel build successfully with `uv build`
- The two previously skipped GitLab auth tests are unskipped and now exercise the real failure path, so the suite has no skips
- New coverage for effective GitHub permission lookup and caching (including 404 denial), ignore authorization and its three modes, bare-name ignore telemetry, malformed comment-table rows, escaping of hostile manifest paths in both the attribute and list contexts, a render-then-parse round trip over every hostile path, change-type preservation through artifact conversion, full-scan output gating and accurate console summaries, namespaced license PURLs, and pull request context resolution

## Upgrade notes

Minor rather than patch: this changes the behavior of existing flags rather than only fixing them.

A pipeline that relied on a default-branch push failing the build on new alerts will now pass; use a pull request pipeline, or `--enable-diff` without `--scm`, to keep a blocking comparison. A pipeline that relied on ignore commands from a commenter without write access will start reporting those alerts. Supply a GitHub token with repository metadata access or a GitLab token that can read project membership to verify permissions; `--ignore-authorization strict` rejects commands when access cannot be determined, `off` restores the previous behavior outright, and `--disable-ignore` turns comment-driven ignores off entirely.

---

Fixes: CE-426, CE-459
Refs: CE-94, CE-215, CE-337, CE-376, CE-424, CE-441

---

> [!NOTE]
> **High Risk**
> Changes CI blocking semantics on SCM branch pipelines, tightens who can suppress alerts via PR comments, and alters HTML/ignore parsing—security-sensitive behavior with broad pipeline impact.
> 
> **Overview**
> **Socket CLI 2.9.0** reshapes how SCM-integrated CI runs choose scan type, associate work with the Dashboard, and honor PR ignore comments—plus several security and output fixes.
> 
> With **`--scm github` or `--scm gitlab`**, only PR/MR events run **diff scans**; default-branch and other non-PR pipelines run **full scans** and no longer block the build (same idea as `--disable-blocking`), even if `--enable-diff` is set. Full scans still fetch packages for **`--generate-license` / FOSSA** output, and console-only full scans say findings were not fetched instead of **“No issues found.”** **`--pr-number`** auto-detects in GitHub Actions, GitLab CI, and Azure Pipelines; Buildkite/`false` values normalize to no PR; diff scans get an **`external_href`** PR/MR URL (with **`on_duplicate=update`** on re-runs). **`--scm`** implies matching **`--integration`** when integration is omitted.
> 
> **`@SocketSecurity ignore`** now requires **write access** (GitHub collaborator permission API with caching; GitLab project membership), controlled by new **`--ignore-authorization`** (`enforce` / `strict` / `off`). Ignore matching is aligned across comment push flows, **ecosystem-qualified** instructions, scoped npm names, legacy table parsing, and telemetry. PR/MR comments show **patched versions**, correct dependency **added/updated/removed/replaced** badges, provider-neutral copy, and **HTML escaping** for repo paths and API alert text.
> 
> **GitLab 401 auth fallback** works again via **`APIFailure`** carrying HTTP status and subclassing the SDK exception. **License PURL** requests preserve package namespaces for scoped/npm and Maven artifacts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ce6d48324684c42a50ae21f05ba8e69bd46bb95. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -1329,7 +1329,8 @@ class Core:
     def get_diff_scan_artifacts(
             self,
             head_full_scan_id: str,
-            new_full_scan_id: str
+            new_full_scan_id: str,
+            external_href: Optional[str] = None
     ) -> DiffArtifacts:
         """Compare two full scans via the diff-scans endpoints, polling for the result.
 
@@ -1352,6 +1353,8 @@ class Core:
         Args:
             head_full_scan_id: The before/base full scan ID
             new_full_scan_id: The after/head full scan ID
+            external_href: Optional pull request or merge request URL to associate
+                with the diff scan in the Socket Dashboard
 
         Returns:
             DiffArtifacts with the added/removed/unchanged/replaced/updated lists
@@ -1361,6 +1364,8 @@ class Core:
             "after": new_full_scan_id,
             "description": f"Socket Security CLI v{__version__} scan comparison",
         }
+        if external_href:
+            create_params["external_href"] = external_href
         try:
             result = self.sdk.diffscans.create_from_ids(self.config.org_slug, create_params)
             diff_scan = result.get("diff_scan") or {}
@@ -1444,7 +1449,8 @@ class Core:
             self,
             head_full_scan_id: str,
             new_full_scan_id: str,
-            include_license_details: bool = False
+            include_license_details: bool = False,
+            external_href: Optional[str] = None
     ) -> Tuple[Dict[str, Package], Dict[str, Package], Dict[str, Package]]:
         """
         Get packages that were added and removed between scans.
@@ -1477,6 +1483,8 @@ class Core:
                 is retained as an explicit override seam, not wired to the
                 ``--exclude-license-details`` user flag (which still governs the
                 human-facing dashboard report URL).
+            external_href: Optional pull request or merge request URL to associate
+                with the primary diff-scan resource
 
         Returns:
             Tuple of (added_packages, removed_packages) dictionaries
@@ -1488,7 +1496,8 @@ class Core:
         try:
             diff_artifacts = self.get_diff_scan_artifacts(
                 head_full_scan_id,
-                new_full_scan_id
+                new_full_scan_id,
+                external_href=external_href,
             )
         except Exception as error:
             # SDK error messages can span many lines (path + response headers); the
@@ -1592,7 +1601,8 @@ class Core:
             save_files_list_path: Optional[str] = None,
             save_manifest_tar_path: Optional[str] = None,
             base_paths: Optional[List[str]] = None,
-            explicit_files: Optional[List[str]] = None
+            explicit_files: Optional[List[str]] = None,
+            external_href: Optional[str] = None
     ) -> Diff:
         """Create a new diff using the Socket SDK.
 
@@ -1604,6 +1614,8 @@ class Core:
             save_manifest_tar_path: Optional path to save manifest files tar.gz archive
             base_paths: List of base paths for the scan (optional)
             explicit_files: Optional list of explicit files to use instead of discovering files
+            external_href: Optional pull request or merge request URL to associate
+                with the diff scan
         """
         log.debug(f"starting create_new_diff with no_change: {no_change}")
         if no_change:
@@ -1728,7 +1740,8 @@ class Core:
         ) = self.get_added_and_removed_packages(
             head_full_scan_id,
             new_full_scan.id,
-            include_license_details=False
+            include_license_details=False,
+            external_href=external_href,
         )
 
         # Separate unchanged packages from added/removed for --strict-blocking support
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -2400,7 +2400,10 @@ class Core:
     @staticmethod
     def add_purl_capabilities(diff: Diff) -> None:
         """
-        Adds capability information to each package in the diff's new_packages list.
+        Adds capability information to the diff's added and updated packages.
+
+        Both lists are walked because an updated package is still newly present at
+        its new version, so its capabilities are as relevant as an added one's.
 
         Args:
             diff: Diff object to update with capability information
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -2214,16 +2214,22 @@ class Core:
         alerts_in_removed_packages: Dict[str, List[Issue]] = {}
         alerts_in_unchanged_packages: Dict[str, List[Issue]] = {}
 
-        seen_new_packages = set()
-        seen_removed_packages = set()
+        seen_packages = {
+            "added": set(),
+            "updated": set(),
+            "removed": set(),
+            "replaced": set(),
+        }
 
         for package_id, package in added_packages.items():
             purl = self.create_purl(package_id, added_packages)
             base_purl = f"{purl.ecosystem}/{purl.name}@{purl.version}"
 
-            if (not direct_only or package.direct) and base_purl not in seen_new_packages:
-                diff.new_packages.append(purl)
-                seen_new_packages.add(base_purl)
+            change_type = "updated" if package.diffType == "updated" else "added"
+            target = diff.updated_packages if change_type == "updated" else diff.new_packages
+            if (not direct_only or package.direct) and base_purl not in seen_packages[change_type]:
+                target.append(purl)
+                seen_packages[change_type].add(base_purl)
 
             self.add_package_alerts_to_collection(
                 package=package,
@@ -2235,9 +2241,11 @@ class Core:
             purl = self.create_purl(package_id, removed_packages)
             base_purl = f"{purl.ecosystem}/{purl.name}@{purl.version}"
 
-            if (not direct_only or package.direct) and base_purl not in seen_removed_packages:
-                diff.removed_packages.append(purl)
-                seen_removed_packages.add(base_purl)
+            change_type = "replaced" if package.diffType == "replaced" else "removed"
+            target = diff.replaced_packages if change_type == "replaced" else diff.removed_packages
+            if (not direct_only or package.direct) and base_purl not in seen_packages[change_type]:
+                target.append(purl)
+                seen_packages[change_type].add(base_purl)
 
             self.add_package_alerts_to_collection(
                 package=package,
@@ -2367,18 +2375,16 @@ class Core:
         Args:
             diff: Diff object to update with capability information
         """
-        new_packages = []
-        for purl in diff.new_packages:
-            if purl.id in diff.new_capabilities:
-                new_purl = Purl(
-                    **{**purl.__dict__,
-                    "capabilities": diff.new_capabilities[purl.id]}
-                )
-                new_packages.append(new_purl)
-            else:
-                new_packages.append(purl)
-
-        diff.new_packages = new_packages
+        for attribute in ("new_packages", "updated_packages"):
+            packages = []
+            for purl in getattr(diff, attribute):
+                if purl.id in diff.new_capabilities:
+                    purl = Purl(
+                        **{**purl.__dict__,
+                        "capabilities": diff.new_capabilities[purl.id]}
+                    )
+                packages.append(purl)
+            setattr(diff, attribute, packages)
 
     def add_package_alerts_to_collection(self, package: Package, alerts_collection: dict, packages: dict) -> dict:
         """
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -1677,6 +1677,14 @@ class Core:
         }
         if external_href:
             create_params["external_href"] = external_href
+            # external_href is only honored while a diff scan is being created,
+            # so re-running a comparison over an already-compared scan pair
+            # would otherwise leave the Dashboard report with no link back to
+            # the pull request. on_duplicate=update applies the link to the
+            # existing resource and answers 200 with the same {"diff_scan": ...}
+            # envelope as a create. Notably it is not on_duplicate=redirect,
+            # whose 302 the SDK follows into a GET without cached=true.
+            create_params["on_duplicate"] = "update"
         try:
             result = self.sdk.diffscans.create_from_ids(self.config.org_slug, create_params)
             diff_scan = result.get("diff_scan") or {}
@@ -1685,11 +1693,14 @@ class Core:
             if error.status_code != 409:
                 raise
 
-            # Do not use on_duplicate=redirect here. The SDK follows that 302
-            # automatically with a GET that lacks cached=true, which can leave
-            # the connection idle while an existing diff scan is still computing.
-            # Resolve the duplicate resource explicitly so every result fetch
-            # continues through the bounded cached polling path below.
+            # Reached without on_duplicate=update (no pull request context to
+            # attach) and against deployments that predate it and still answer
+            # 409 regardless. Do not switch this to on_duplicate=redirect: the
+            # SDK follows that 302 automatically with a GET that lacks
+            # cached=true, which can leave the connection idle while an existing
+            # diff scan is still computing. Resolve the duplicate resource
+            # explicitly so every result fetch continues through the bounded
+            # cached polling path below.
             existing = self.sdk.diffscans.list(
                 self.config.org_slug,
                 params={
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -1708,12 +1708,9 @@ class Core:
         if external_href:
             create_params["external_href"] = external_href
             # external_href is only honored while a diff scan is being created,
-            # so re-running a comparison over an already-compared scan pair
-            # would otherwise leave the Dashboard report with no link back to
-            # the pull request. on_duplicate=update applies the link to the
-            # existing resource and answers 200 with the same {"diff_scan": ...}
-            # envelope as a create. Notably it is not on_duplicate=redirect,
-            # whose 302 the SDK follows into a GET without cached=true.
+            # so a re-run over an already-compared scan pair needs
+            # on_duplicate=update to apply the link to the existing resource. It
+            # answers 200 with the same {"diff_scan": ...} envelope as a create.
             create_params["on_duplicate"] = "update"
         try:
             result = self.sdk.diffscans.create_from_ids(self.config.org_slug, create_params)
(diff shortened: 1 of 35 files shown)


socketsecurity/core/scm_comments.py
```diff
@@ -66,20 +66,41 @@ def get_ignore_options(comments: dict) -> [bool, list]:
...
+        ``ignore @types/node@*`` would also silently ignore alerts for a package
+        literally named ``node``.
+        """
+        package_names = {pkg_name}
+        if pkg_type:
+            package_names.add(f"{pkg_type}/{pkg_name}")
+        target_names = {name}
+        if not pkg_type and "/" in name and not name.startswith("@"):
+            target_names.add(name.split("/", 1)[1])
+        return bool(package_names & target_names) and (pkg_version == version or version == "*")
```
flowstate, line 103: This function has no way to know whether the commenter is authorized to ignore an alert -- `Comment.author_association` exists on the dataclass but is never checked anywhere in the call path. This PR is what makes the bare-name fallback here actually functional (it used to throw on any `@`-containing name), so it's the right spot to add the check rather than deferring it.

Also: `package_names` always includes the bare `pkg_name` regardless of `pkg_type`, so a bare ignore command matches across every ecosystem with that name+version, not just the one that was actually alerted.

socketsecurity/core/pull_request.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _github_url(number: int, repo: Optional[str], env: Mapping[str, str]) -> Optional[str]:
+    remote_host, remote_path = parse_git_remote(env.get("BUILDKITE_REPO"))
+    # config.repo is only ever a bare repository name, so it cannot produce a
+    # slug on its own; it is kept last for callers that pass a full owner/repo.
+    repository = env.get("GITHUB_REPOSITORY") or remote_path or repo
+    if not repository or "/" not in repository:
+        return None
+    server = env.get("GITHUB_SERVER_URL") or (f"https://{remote_host}" if remote_host else "")
```
flowstate, line 55: `server` is taken directly from `GITHUB_SERVER_URL` with no scheme/netloc validation, unlike `_repository_url` just above which checks `parsed.scheme in ("http", "https") and parsed.netloc` for the sibling env vars. Not reachable through standard GitHub Actions (runner-set), but this value becomes `external_href` sent to the API, so routing it through the same validator would be cheap defense-in-depth.

socketsecurity/core/pull_request.py
```diff
@@ -0,0 +1,125 @@
...
+    server = (server or "https://github.com").rstrip("/")
+    return f"{server}/{repository.strip('/')}/pull/{number}"
+
+
+def _gitlab_url(number: int, repo: Optional[str], env: Mapping[str, str]) -> Optional[str]:
+    project_url = _repository_url(env.get("CI_PROJECT_URL"))
+    if not project_url:
+        remote_host, remote_path = parse_git_remote(env.get("BUILDKITE_REPO"))
+        project_path = env.get("CI_PROJECT_PATH") or remote_path or repo
+        server = env.get("CI_SERVER_URL") or (f"https://{remote_host}" if remote_host else "")
```
flowstate, line 65: Same gap as `GITHUB_SERVER_URL` above -- `CI_SERVER_URL` is used raw here too.

flowstate (changes requested):
[agent] Reviewed via two independent passes (security-focused and quality-focused) over the diff against main, plus manual verification of the two most significant findings against the actual source.

Requesting changes on one high-severity issue and one adjacent bug this PR doesn't touch but sits right next to.

**Ignore-comment authorization.** `Comment.author_association` is defined on the `Comment` dataclass but never read anywhere in the codebase -- there's no check that the person posting `@SocketSecurity ignore <pkg>@<version>` actually has write access to the repo. That gap predates this PR, but this PR is what makes it exploitable in practice: bare (non-ecosystem-qualified) ignore commands were actually broken before (the old parser threw on any scoped package name and the error was swallowed), and the new scope-disambiguation logic here makes that path work again. See the inline comment on `is_ignore` in `scm_comments.py`.

**Adjacent crash bug, not touched by this PR.** `process_original_security_comment` (the legacy comment-table format) still does `pkg_name, pkg_version = details.split("@")` on a scoped package name, which unpacks into 3 parts and raises an uncaught `ValueError` -- the exact bug class this PR fixes in `process_updated_security_comment` via `rsplit("@", 1)`, just left unfixed in its sibling function. `socketsecurity/core/scm_comments.py:155`. Worth fixing in the same pass since it's the same bug, in the same file, one function over -- couldn't attach this as an inline comment since the line isn't part of this diff.

Also flagging, not blocking, both left as inline comments: `_github_url`/`_gitlab_url` in `pull_request.py` build the dashboard-link host from `GITHUB_SERVER_URL`/`CI_SERVER_URL` without the scheme/netloc validation `_repository_url` already applies to the sibling env vars in the same file.

One tiny doc fix, not on a changed line so no inline comment for it: `add_purl_capabilities`'s docstring in `socketsecurity/core/__init__.py:2346` still says this only touches `new_packages`, but the loop now also processes `updated_packages`.

Everything else checked out: the diffType classification, the `create_scm_scan` full-scan/blocking refactor, the PR-context resolution precedence, and the argparse-suppress plumbing were all traced by hand and are well covered by the new tests.

socketsecurity/core/scm_comments.py
lelia: Fixed in 977a8fb — good catch, and you're right that this was the spot to do it rather than defer.

The gate went into `check_for_socket_comments` rather than `is_ignore`, since that's the one place every consumer passes through: `remove_alerts`, `process_security_comment`, and both ignore-telemetry loops in `socketcli.py` all read the already-bucketed dict. A rejected command is therefore also absent from the telemetry, which should record what was acted on.

GitHub is free and definitive — `author_association` is already in the comment payload, so `OWNER`/`MEMBER`/`COLLABORATOR` only, and a missing value is treated as unauthorized rather than trusted.

GitLab needed more thought. Notes carry no permission field, and my first attempt (`GET /projects/:id/members/all/:user_id`) turned out to be unsafe here: `CliClient.request` funnels every HTTP error through `raise_for_status()` into `APIFailure` **without** a status code, so a 404 meaning "not a member" — exactly the outsider case this guards — is indistinguishable from a token that can't read the endpoint, and would have to fail open. Switched to `members/all`, which answers non-membership with a 200 and an absent id, so the answer is definitive whenever the list is readable. Developer (30) or above, fetched once per run and only when an ignore command is actually present.

One deliberate asymmetry worth flagging: when the GitLab list genuinely can't be read — a `CI_JOB_TOKEN` generally can't, and that's what the shipped `workflows/gitlab-ci.yml` template uses — the command is honored and a warning names the author. Failing closed there would break every templated GitLab pipeline that relies on ignore commands. Documented under "Who can ignore an alert" in `cli-reference.md` with the token needed to get enforcement. Happy to flip it to fail-closed if you'd rather take the breakage.

On the second half — bare names matching across ecosystems — that one I left as-is, deliberately. Pre-2.9.0 the generated command was `ignore {pkg_name}@{version}` with no ecosystem at all, so a bare command sitting in an existing comment genuinely carries no ecosystem information; name+version is the only thing available to match on. Narrowing it would break the round-trip with older comments that this PR set out to fix. The version still has to match exactly (or be `*`), and with the authorization check in place the blast radius is bounded to people who could have ignored the alert anyway. I did fix the adjacent real over-match in d49facb, where a leading npm scope was being stripped as though it were an ecosystem, so `ignore @types/node@*` no longer also matches a package named `node`.

22 new tests in `tests/unit/test_ignore_authorization.py`, including `ignore-all` from an outsider (the more powerful command, and one that did work before this PR).

socketsecurity/core/pull_request.py
lelia: Fixed in f16525f. Pulled the scheme/netloc check out of `_repository_url` into a shared `_http_url`, and routed `GITHUB_SERVER_URL` through it alongside the `https://{remote_host}` fallback derived from `BUILDKITE_REPO`.

Agreed it's defense in depth rather than a live hole — the runner sets this — but as you say it ends up as `external_href` on the diff scan, so it's cheap to validate at the boundary.

An unusable value falls back to `https://github.com` rather than dropping the link, since GitHub has a well-known public host. Covered by a parametrized test over `javascript:`, a bare string, `ftp://`, a scheme with no host, and empty, plus one asserting self-hosted GHE URLs still work.

socketsecurity/core/pull_request.py
lelia: Same fix in f16525f — `CI_SERVER_URL` goes through the same `_http_url` validator.

One difference from the GitHub side: GitLab has no public default host to fall back to, so an unusable `CI_SERVER_URL` drops the merge request link entirely rather than guessing gitlab.com. The scan still carries its MR number, it just doesn't get a Dashboard link. Tested both ways, including that a self-hosted `http://gitlab.internal` with a nested subgroup path still composes correctly.

lelia:
@flowstate thanks — all four items addressed, plus the rebase. Replies are inline on the three threads; summarizing the two that had no anchor line:

**Adjacent crash bug in `process_original_security_comment`** — fixed in 63e938d. You were right that it's the same bug one function over: `details.split("@")` unpacks into three parts on a scoped name. Now `rsplit("@", 1)`, matching `process_updated_security_comment`.

While in there I also stopped it pre-concatenating the ecosystem onto `pkg_name` and instead pass it through as `pkg_type`. That fixed a second asymmetry you didn't flag: because `pkg_name` arrived pre-qualified as `npm/@scope/pkg`, the legacy path matched only the ecosystem-qualified ignore command and silently ignored the bare form, while the current format accepted both. Both formats now accept both forms.

**`add_purl_capabilities` docstring** — fixed in d7b01b0.

---

Two other things in this push:

**Version is now 2.9.0, not 2.8.0.** 2.8.0 and 2.8.1 both shipped from main while this was open, so the original bump was dead. Minor rather than patch per the repo's semver standard, since this changes the behavior of existing flags rather than only fixing them: `--pr-number` gains auto-detection, `--scm github|gitlab` implies `--integration`, and SCM branch pipelines switch from diff scans to full scans and stop returning a blocking exit code.

**Rebased onto `6864479` (v2.8.1).** The branch had been conflicting, which meant GitHub couldn't build a merge ref and the `pull_request`-triggered workflows — Unit Tests, E2E, Package Check, Version Check, dependency-review — weren't running at all; only the head-commit checks were. Two conflicts: an import block re-sorted by #342's ruff autofixes, and the changelog. The changelog one is worth noting because it does not surface as a conflict marker — our section was headed `## 2.8.0` and main has since shipped a real `## 2.8.0`, so git resolves the duplicate header in main's favour and silently deletes the entries. That happened on a previous rebase of this branch. Renaming to `## 2.9.0` first avoided it, and I verified the section afterwards.

642 passing, 2 skipped. `ruff check` clean repo-wide against the job #342 added.

lelia:
bugbot run

socketsecurity/core/__init__.py
```diff
@@ -1980,7 +2030,8 @@ def create_new_diff(
             save_files_list_path: Optional[str] = None,
             save_manifest_tar_path: Optional[str] = None,
             base_paths: Optional[List[str]] = None,
-            explicit_files: Optional[List[str]] = None
```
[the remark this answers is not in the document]
lelia: Checked this one carefully — the central claim doesn't hold, but it surfaced a real test gap, so thank you for it.

**`diffType` is assigned in the production path.** `DiffArtifact.diffType` is a required field on the SDK dataclass (`diffType=DiffType(data["diffType"])`), and `Package.from_diff_artifact` reads it in both response shapes:

- **Flattened** (`topLevelAncestors` present, no refs — what the API sends): `diffType=diff_type`, after unwrapping the enum via `hasattr(diff_type, "value")`.
- **Ref-shaped** (head/base refs): `diffType=ref.get("diffType", diff_type)`. `SocketArtifactLink` has no `diffType` field, so the key is absent and the fallback applies.

Traced end to end on real `DiffArtifact` objects for all four change types across both shapes:

```
  added     flattened='added'     ref-shaped='added'
  updated   flattened='updated'   ref-shaped='updated'
  removed   flattened='removed'   ref-shaped='removed'
  replaced  flattened='replaced'  ref-shaped='replaced'
```

and the classification `create_diff_report` performs on those values routes each to the right bucket.

**Where you were right:** the test guarding this assigned `diffType` by hand (`added["dp3"].diffType = "updated"`), so it exercised the bucketing but not the plumbing — it would have passed whether or not the conversion populated the field. That's exactly the blind spot your finding describes, even though the conversion happens to be correct today.

Fixed in `7a986c8` with `test_change_type_survives_artifact_conversion`, parametrized over all four change types and both response shapes, running real `DiffArtifact` objects through the real conversion. Verified it has teeth by mutating `from_diff_artifact` to drop the field: all 8 new cases fail, while the pre-existing stubbed test still passes.

Not marking this resolved unilaterally in case you're seeing something I'm not — if the concern is a specific code path I haven't covered, point me at it.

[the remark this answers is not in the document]
lelia: Follow-up: the test fix is now on the branch head (`7a986c8`), so this is ready for a re-check.

To restate the verdict for the re-run: `diffType` **is** assigned in the production path, in both response shapes — so the reported defect (updates shown as additions, replacements as removals) does not occur. What was genuinely missing was a test that proved it, since the existing one assigned the field by hand. That test now exists and is parametrized over all four change types × both shapes.

Re-running Bugbot below. If it still flags this, I'd want the specific call path it thinks drops the field, because the four traces above cover every branch of `Package.from_diff_artifact`.

lelia:
bugbot run

socketsecurity/core/__init__.py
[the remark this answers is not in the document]
lelia: Re-run result: Bugbot completed on `7a986c8` in 2m46s with a **passing** check and raised no findings — including not re-raising this one against the current code.

Recording the outcome precisely rather than overclaiming: the run came back clean, but Bugbot did not explicitly retract this comment, and the original is still anchored here. So this is "not reproduced on re-run" rather than a formal retraction.

Substance is unchanged from the trace above — `diffType` is assigned in both response shapes, and the gap was the test, which now covers all four change types across both shapes. Leaving the thread open for a human to close.

lelia:
bugbot run

lelia:
bugbot run

flowstate (approved):
[agent] Second pass looks good. Traced the ignore-authorization gate (GitHub permission lookup + caching + 404-as-denial, GitLab membership lookup + pagination bound) and the scoped-name matching logic in `is_ignore` by hand across both comment formats — correct. The two non-blocking notes from the first pass (URL validation, docstring) are also addressed. Ran the suite locally: 680 passed (remaining 8 errors are a local sandbox git-init artifact, not a repo issue) and `ruff check` is clean. Approving.

PR #321, opened 2026-08-28
Don't render `.socket.facts.json` placeholder as blocking package

lelia:
## Problem

When a scan finds no supported manifest files, the CLI uploads a placeholder named `.socket.facts.json` so the upload passes the API's supported-filename validator. That placeholder was zero bytes, which the API cannot parse as a facts document — it responds by adding a synthetic `generic/invalid-socket-facts@1.0.0` artifact carrying a blocking alert.

The CLI read that back as a real dependency, so the run reported a new blocking package with no manifest file and no introducing dependency, exited non-zero, and on pull requests posted a security comment that a developer had no way to act on.

## Changes

- **Placeholder is now valid.** It contains an empty but well-formed facts document, so the scan is genuinely empty instead of triggering a parse error.
- **Placeholders no longer share a path.** Each gets its own temp directory, removed at exit. The previous fixed `$TMPDIR/.socket.facts.json` meant two runs sharing a temp directory could remove each other's file mid-upload.
- **The marker is filtered out regardless of origin.** `generic/invalid-socket-facts` is dropped from full scan artifacts and from every diff bucket (added/removed/unchanged/replaced/updated), with a warning logged in its place. Filtering happens before the artifact-count log lines so the counts describe what the CLI actually reports on. With no new alerts, the existing gate in `socketcli.py` suppresses the security comment on its own.

## Testing

`tests/core/test_invalid_facts_marker.py` (13 tests). The eight behavioral ones were confirmed to fail against the pre-fix code. Full suite: 492 passed.

## Note

That the backend accepts the new placeholder payload is not verified against a live API — there is no no-manifest scenario in the e2e workflow. If the schema wants more, the marker filtering still prevents the block and the comment, but the placeholder would keep tripping the parser. Worth a preview build against a repository with no manifests to confirm the resulting scan carries no `invalid-socket-facts` artifact.

Fixes CE-422

---

> [!NOTE]
> **Medium Risk**
> Changes core scan/diff artifact handling and empty-scan upload behavior, which directly affects blocking exits and PR comments, though the logic is narrow and well-tested.
> 
> **Overview**
> Fixes **CE-422**, where scans with no manifest files could fail CI and post unusable PR security comments.
> 
> **Placeholder uploads** no longer write a zero-byte `.socket.facts.json` (which the API could not parse). They now upload a minimal valid document (`{"components": []}`), and each run gets its own temp directory (with exit-time cleanup) instead of sharing a fixed `$TMPDIR/.socket.facts.json` path that concurrent jobs could clobber.
> 
> **API diagnostic artifact** `generic/invalid-socket-facts` is treated as a parse-failure signal, not a dependency: it is stripped from full-scan SBOM artifacts and from every diff bucket before package/alert assembly and count logging, with a **warning** when present. Runs no longer treat it as a new blocking package or PR comment fodder.
> 
> Release **2.6.9** with changelog and focused tests in `test_invalid_facts_marker.py`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01b89e48828b7954b792607bbfd1be9ad86ed5e3. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/socketsecurity/core/__init__.py
+++ b/socketsecurity/core/__init__.py
@@ -1,3 +1,4 @@
+import atexit
 import copy
 import fnmatch
 import importlib
@@ -6,6 +7,7 @@ import logging
 import os
 import random
 import re
+import shutil
 import sys
 import tarfile
 import tempfile
@@ -69,6 +71,16 @@ SOCKET_FACTS_BROTLI_LGWIN = 24
 # Stream the facts file in 1 MiB chunks so large files aren't held fully in memory.
 SOCKET_FACTS_BROTLI_CHUNK_SIZE = 1024 * 1024
 
+# Placeholder facts document (see empty_head_scan_file). A zero-byte file does not parse,
+# and the API answers an unparseable facts file with the marker artifact below.
+SOCKET_FACTS_EMPTY_DOCUMENT = '{"components": []}'
+
+# Synthetic artifact the API adds when an uploaded ``.socket.facts.json`` could not be
+# parsed. A diagnostic, not a dependency, so it is dropped from scan results and reported
+# as a warning instead.
+INVALID_FACTS_MARKER_TYPE = "generic"
+INVALID_FACTS_MARKER_NAME = "invalid-socket-facts"
+
 # Full application reachability finalize retry policy. The finalize call links the reachability
 # scan to the full scan and can fail transiently (network/API blips); a few backoff retries make it robust.
 TIER1_FINALIZE_MAX_ATTEMPTS = 3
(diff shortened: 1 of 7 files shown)


socketsecurity/core/__init__.py
```diff
@@ -69,6 +71,19 @@
...
+# Minimal well-formed facts document used for placeholder uploads (see empty_head_scan_file).
+# A zero-byte ``.socket.facts.json`` is not parseable, and the API answers an unparseable
+# facts file by synthesising the marker artifact below.
+SOCKET_FACTS_EMPTY_DOCUMENT = '{"components": []}'
+
+# Synthetic artifact the Socket API adds to a full scan when an uploaded
+# ``.socket.facts.json`` could not be parsed. It is a diagnostic marker rather than a real
+# dependency: it has no manifest file and no introducing package, so the blocking alert it
+# carries is not actionable by a developer, and a PR comment about it is pure noise
+# (CE-422). Drop it from scan results and surface the parse failure as a warning instead.
```
flowstate: There are 8 spots where the ticket is directly mentioned in comments. That looks to me to be an LLM agent special, and I've got a skill to clear them out without having to pick through them all. 

DM me if you disagree with this but I think that we just ship this version, and I'll do a PR tomorrow that adds the skill and scrubs these out of the main branch (assuming this is merged).

flowstate (approved):
See my note about comment nits, but the important stuff lgtm

socketsecurity/core/__init__.py
lelia: Good catch, and agreed: scrubbed all 8 in this PR rather than leaving them for you. Also trimmed the surrounding comments, which were padded out well past what the code needed, and dropped the `Ref:` trailer from the commit message (squash here uses `COMMIT_MESSAGES`, so it would have landed on main). The ticket is now referenced only in the PR description footer.

PR #322, opened 2026-08-28
Fix pull request comment rendering and disable flags

lelia:
## Summary

Two fixes to the pull request comment the CLI posts:

1. **Rendering artifacts.** Security comments could render orphaned `</blockquote></details>` tags as a literal code block, and could leave the "Caution" banner sitting above an alerts table with no rows. A whitespace-only line closes a CommonMark HTML block; optional sections that rendered as empty left one behind inside the table, and because the closing tags after it are indented four or more spaces they were rendered as an indented code block rather than markup. The empty table appeared separately, when a comment's alerts were all resolved by a later commit or ignored by comment.
2. **Comment disable flags.** `--disable-security-issue` and `--disable-overview` were checked only after testing whether a comment of that type already existed, so they suppressed the first post and then updated that comment on every later run.

## Changes

- Generated comment markup no longer contains blank lines and stays under the indentation that starts a code block.
- Alert descriptions, suggestions and license findings are collapsed onto a single line, so multi-line API text cannot break the table either.
- When no alerts are left to report, the comment body is replaced with a short confirmation instead of an empty table. The comment marker is preserved, so a commit that reintroduces an alert updates the same comment rather than posting a second one.
- `@SocketSecurity ignore-all` now applies to comments in the pre-2.0.55 Markdown table format. The check was made once per ignore command, and an ignore-all comment produces none, so no rows were removed.
- The disable flags now mean the CLI does not manage that comment at all. An existing comment is left untouched rather than rewritten, since a body claiming no alerts would be inaccurate when reporting is merely switched off.

Bumps to 2.7.0 rather than a patch, since the disable flags change behavior.

## Testing

- `tests/unit/test_pr_comment_rendering.py` covers both rendering artifacts, the ignore round trips, and the collapsed body. `tests/unit/test_socketcli.py` covers the comment write decision, which previously had no harness.
- Every generated body was rendered through GitHub's `/markdown` API to confirm the tags are gone; 423 unit and 87 core tests pass.

Fixes CE-381
Fixes CE-427

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,18 @@
   once per ignore command, and an ignore-all comment produces none, so no rows
   were removed.
 
+### Fixed: `--disable-security-issue` and `--disable-overview` now suppress the comment entirely
+
+- Both flags were checked only after testing whether a comment of that type was
+  already on the pull request, so they suppressed the first post and then
+  updated that comment on every later run. `--disable-security-issue` in
+  particular kept refreshing an existing comment with the full alerts table.
+- The flags now mean the CLI does not manage that comment at all. An existing
+  comment is left untouched rather than being rewritten, since a body claiming
+  no alerts would be inaccurate when reporting is merely switched off.
+- The decision moved into `should_write_comment()` so it is covered directly by
+  tests.
+
 ## 2.6.11
 
 ### Changed: bump pinned @coana-tech/cli to 15.10.32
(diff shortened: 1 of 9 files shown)


flowstate (approved):
LGTM!

oh and we should add a lint:fix via ruff somewhere in our automation, main is carrying around like 75 formatting issues, nothing major

PR #341, opened 2026-09-04
Make tar_hash required for reachability analysis

BarrensZeppelin:
`run_reachability_analysis()` always receives a manifest tar hash from its only caller, so `tar_hash` no longer defaults to `None`.

- Drop the `Optional[str] = None` default in favour of a required `str`.
- Drop the conditional that skipped `--run-without-docker` / `--manifests-tar-hash` when no hash was present; those flags are now always passed.
- Simplify manifest discovery in `socketcli.py` into a comprehension.
- Pass `tar_hash` at the test call sites.

No test assertions changed: nothing in the suite covered `--manifests-tar-hash` or `--run-without-docker`, so the removed branch was untested in either direction.

---

> [!NOTE]
> **Low Risk**
> Contract tightening on an internal API with a single caller; behavior for the supported CLI flow is unchanged aside from always forwarding manifest hash flags to Coana.
> 
> **Overview**
> **Reachability analysis now always requires a manifest upload tar hash** and always passes it to Coana via `--run-without-docker` and `--manifests-tar-hash`.
> 
> `run_reachability_analysis()` takes a required `tar_hash: str` instead of optional `None`, and the branch that omitted those flags when no hash was present is removed. That matches the CLI path, which only invokes the analyzer after `upload_manifest_files` succeeds.
> 
> In `socketcli.py`, manifest discovery is collapsed into one comprehension (still excluding `.socket.facts.json`), and `assert org_slug` is added before upload. Unit tests pass a fixed `tar_hash` at every call site; assertions are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b7e87a8d5c37ef0c04cd27e3612ff94d20703c0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/socketsecurity/core/tools/reachability.py
+++ b/socketsecurity/core/tools/reachability.py
@@ -77,7 +77,7 @@ class ReachabilityAnalyzer:
         self,
         org_slug: str,
         target_directory: str,
-        tar_hash: Optional[str] = None,
+        tar_hash: str,
         output_path: str = ".socket.facts.json",
         timeout: Optional[str] = None,
         memory_limit: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


mtorp (approved):
LGTM ✅

PR #345, opened 2026-09-07
Install the pinned reachability engine in the Docker image

lelia:
## Problem

`Dockerfile` installs `@coana-tech/cli` unpinned, while the launcher asks npx for the version in `DEFAULT_COANA_CLI_VERSION`. npx reuses the image's global install only when the versions match; on a mismatch it downloads the engine again on every scan.

Measured in the published image:

| Requested version | Launcher | npm cache delta |
|---|---|---|
| matches the global install | 3s | 0 MB |
| mismatched | 9s | +119 MB |

The two agree only when a release rebuild happens to follow a pin bump. Nothing enforces it, and the failure mode is silent — no error, no test failure, just slower scans and extra egress. The pin moved from `15.10.36` to `15.10.51` while this PR was open, across fifteen bumps.

## Change

- The Dockerfile reads the pinned version out of `reachability.py` and installs exactly that, failing the build if the value cannot be read. The pin stays bumped in one place, so the image and the runtime cannot drift.
- The coana install moves out of the toolchain `RUN`. A pin bump now rebuilds a 353 MB layer instead of the 2.3 GB combined layer — relevant given how often the pin moves.
- Marking a release stable builds that version from its own tag. The image installs `socketsecurity==inputs.version` from PyPI, so building from the default branch paired an old wheel with a newer build recipe and pin.
- `tests/unit/test_dockerfile_coana_pin.py` runs the Dockerfile's own extraction expression against the real source and rejects an unpinned install. `Dockerfile` is listed in the Unit Tests `push` paths since the suite now reads it.

## Verification

- Full image build: baked engine, runtime pin and source agree; `socketcli --help` and `coana-cli` both work. Re-verified after rebase against the current pin (`15.10.51`).
- Building against a source file with the constant renamed aborts with `Could not read DEFAULT_COANA_CLI_VERSION from reachability.py` rather than silently installing unpinned.
- After the change npx resolves without downloading — metadata only, no cache growth.
- Both guards fail on their respective mutations and pass when reverted.
- 575 unit tests pass; `ruff check` clean. The extraction expression was checked under both BSD and busybox `sed`.

No CI job builds this Dockerfile on a pull request, so the image verification above is local. e2e-reachability covers the launcher path only.

## Not included

`socket` npm is still installed unpinned and the npm cache is still shipped in the image. Both are tracked separately.

---

Ref: CE-431

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/docker-stable.yml
+++ b/.github/workflows/docker-stable.yml
@@ -15,6 +15,11 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
+          # Build the requested version from its own tag. The image installs
+          # socketsecurity==inputs.version from PyPI, and the Dockerfile reads the pinned
+          # @coana-tech/cli version out of the checked-out source, so building from the
+          # default branch would pair an old wheel with a newer build recipe and pin.
+          ref: v${{ inputs.version }}
           persist-credentials: false
 
       - name: Check if version exists in PyPI
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -12,6 +12,7 @@ on:
       - "tests/core/**/*.py"
       - "pyproject.toml"
       - "uv.lock"
+      - "Dockerfile"
       - ".github/workflows/python-tests.yml"
   pull_request:
     paths:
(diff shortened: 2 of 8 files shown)


lelia:
bugbot run

PR #365, opened 2026-09-21
Bump gitpython to 3.1.62 and soupsieve to 2.9.2

lelia:
## Dependency bumps

Both pins are flagged by `pip-audit` against the current lock.

| Package | From | To | Advisories |
| --- | --- | --- | --- |
| gitpython | 3.1.59 | 3.1.62 | CVE-2026-87817, CVE-2026-87818, CVE-2026-87819 |
| soupsieve (transitive) | 2.8.4 | 2.9.2 | GHSA-gjv8-xp57-g29c, GHSA-j934-xhv5-fg8f |

Supersedes #361.

## Pin casing

`GitPython==` becomes `gitpython==`, the only mixed-case requirement in the project; `uv.lock` already uses the normalized name. Dependabot's uv updater errors on this entry with `dependency_file_content_not_changed`, and the rename is a candidate fix. Inert either way, since PEP 503 names are case-insensitive to pip and uv and this was the only occurrence of the string.

## Audit scheduling

`pip-audit` moves out of the Unit Tests workflow into a new Dependency Audit workflow with a daily cron, keeping the pull request trigger. The audit compares the lockfile against databases that publish continuously, so its result tracks the clock rather than the commit and needs a trigger to match. A failing scheduled run has no pull request to report on, so it opens a tracking issue and closes it once the audit is clean.

## Check triggers

The `paths` filters come off the `pull_request` triggers for Unit Tests and Version Check. A status check behind a `paths` filter produces no check context on a pull request that misses the filter, so those jobs cannot be marked required while the filters are in place. The `push` triggers keep theirs. Each job finishes in under a minute.

## Verification

```
uv lock --locked      exit 0
pytest                717 passed
import smoke          OK
pip-audit --strict    No known vulnerabilities found
ruff check            All checks passed!
```


CHANGELOG.md
```diff
@@ -1,5 +1,40 @@
 # Changelog
 
+## 2.9.5
```
BarrensZeppelin, line 3: Are you sure you want to bump the version?

If we do, I think we should heavily trim the changelog, which is user-facing. 🙂

.github/workflows/python-tests.yml
```diff
@@ -62,10 +60,10 @@ jobs:
...
           print('import smoke OK')
           "
-      - name: 🛡️ pip-audit (known CVEs in the locked deps)
-        run: |
-          uv export --no-hashes --no-emit-project --format requirements-txt > /tmp/req-audit.txt
-          uvx pip-audit --strict --progress-spinner off --disable-pip --no-deps -r /tmp/req-audit.txt
+      # pip-audit used to run here. It moved to dependency-audit.yml, because
+      # its result depends on when it runs rather than on the commit, and
+      # mixing the two meant a third-party advisory publication reported itself
+      # as "Unit Tests failed" while every test passed.
```
BarrensZeppelin: [agent] This one is a tombstone for the step the diff deletes: it records what used to be here, that it moved, and why the move was made. Once this PR merges, nobody reading the workflow needs to know there was ever a `pip-audit` step in it — and the rationale for splitting the two checks is already in the header of `dependency-audit.yml`, where it applies to code that exists. I'd drop the comment entirely.

.github/workflows/dependency-audit.yml
```diff
@@ -0,0 +1,166 @@
...
+  PYTHON_VERSION: "3.12"
+
+# This check is not a function of the code alone. pip-audit compares the locked
+# dependency set against advisory databases that publish continuously, so a
+# green run only describes the moment it ran -- the same commit can fail
+# tomorrow with nothing changed. That is why this runs on a schedule and not
+# only on pull requests: between 2026-09-09 (three GitPython advisories
+# published) and 2026-09-17 (the next push that happened to touch a path the
+# test workflow watches) the repository was shipping a known-vulnerable pin
+# with every check green.
```
BarrensZeppelin: [agent] The first three lines earn their place — that a green run only describes the moment it ran is exactly the non-obvious thing a reader needs in order to not "simplify" the schedule away later.

From `between 2026-09-09` onward it turns into the incident writeup: the dates, what the test workflow's paths filter happened to watch, what was shipping in the meantime. That argues for creating this file, which is a PR-description job, and the CHANGELOG entry already covers it. Trimming to the first four lines (through `not only on pull requests`) keeps everything a future reader can use.

.github/workflows/python-tests.yml
```diff
@@ -13,14 +13,12 @@ on:
       - "pyproject.toml"
       - "uv.lock"
       - ".github/workflows/python-tests.yml"
+  # Deliberately unfiltered, unlike the push trigger above. A required status
+  # check that sits behind a paths filter never reports on a pull request that
+  # misses the filter, and an unreported required check blocks the merge
+  # forever. These jobs finish in well under a minute, so running them on every
+  # pull request costs little and is what makes them safe to require.
```
BarrensZeppelin: [agent] The middle sentence is the real content, and it's worth keeping — a required check behind a paths filter never reports, so the merge blocks forever. That's the thing someone would otherwise undo.

The bookends are aimed at this diff rather than at the file: `unlike the push trigger above` contrasts against something a reader can see two lines up, and `costs little and is what makes them safe to require` is a cost argument for removing the filter. Both read as answers to a question only a reviewer would ask.

BarrensZeppelin (approved):
[agent] Approving. This is a comments-only pass over the three workflow files. Three notes inline, all about comments that speak to a reviewer of this diff rather than to someone reading the workflow later; the first is the one I'd actually act on.

[human] LGTM - thanks! :tada:

CHANGELOG.md
lelia: Yes, the bump is required. 2.9.4 is already on PyPI, so `check_version` fails on `pr_ver <= pypi_ver`, and twine would reject the duplicate at release. The pins are exact and ship in the wheel metadata, so the bumped `gitpython` only reaches installs via a release.

Changelog trimmed to four bullets across the two sections.

.github/workflows/python-tests.yml
lelia: Dropped the comment entirely.

.github/workflows/dependency-audit.yml
lelia: Trimmed past your suggestion — the header is now the two lines of rationale, with the dates and the incident writeup gone:

```
# pip-audit compares the lockfile against advisory databases that publish
# continuously, so the result tracks the clock, not the commit. Hence the cron.
```

.github/workflows/python-tests.yml
lelia: Kept the middle sentence, dropped both bookends:

```
# Unfiltered: a required check behind a paths filter never reports on a
# pull request that misses the filter, which blocks the merge.
```

Same wording now in `version-check.yml` and `dependency-audit.yml`.

